PR #51, opened 2018-08-30
Feature/sdpa 865 bggraphic

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-865

### Changed

* Added backgroundGraphic to PageLayout component.
* Added backgroundGraphic to Page demo.
* Removed backgroundImage from Page demo (was pointing to an undefined variable).
* Moved styles from NUXT main.scss to the PageLayout component for rpl-above-content.
* Moved rpl-above-content__inner to within the rpl-above-content definition (match other implementations).

### Screenshots

mobile-graphic
desktop-graphic


packages/Molecules/Layout/PageLayout.vue
```diff
@@ -59,6 +60,9 @@ export default {
       }
 
       return {}
+    },
+    bgGraphic () {
+      return this.backgroundGraphic ? { 'background-image': `url(${this.backgroundGraphic})` } : null
```
tim-yao: Hi @alan-cole I feel this logic is not needed, because background image will cover the graphic, am I right?
That can save make everything on it's own, not have coupled logic.

alan-cole: If we do this, then a background graphic will need to be defined for every use of PageLayout. (Otherwise we get this)
screen shot 2018-09-03 at 11 39 59 am

Personally, I think it's a better solution if you can choose to use a graphic or not, rather than force the use of a graphic (or risk broken HTML), but this is your call - I will implement as requested.

PR #66, opened 2018-09-12
[SDPA-906] list

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-906

### Changed

1.  Added list component.

### Screenshots

screen shot 2018-09-12 at 2 53 07 pm

### DIFF
--- a/packages/Molecules/List/index.vue
+++ b/packages/Molecules/List/index.vue
@@ -1,0 +1,132 @@
+<template>
+  <div
+    class="rpl-list"
+    :class="{
+      'rpl-list--normal': (size === 'normal'),
+      'rpl-list--large': (size === 'large')
+    }"
+  >
+    <h2 v-if="title" class="rpl-list__title">{{ title }}</h2>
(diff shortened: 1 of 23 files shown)


packages/Molecules/List/index.vue
```diff
@@ -0,0 +1,119 @@
...
+    iconColor: { type: String, default: 'primary' },
+    list: Array
+  },
+  components: {
+    RplLink,
+    RplIcon
+  },
+  methods: {
+    iconSize (item) {
+      return (item.size ? item.size : 1) * this.iconScale
```
alan-cole: An icon size can be set per icon. Icons can also be globally scaled using iconScale on the component.
If an icon size is set to 1.2, then iconScale is set to 2.5, the resulting icon will be (1 * 1.2 * 2.5) = 3.
This means minor adjustments to icon sizes can be made on a per icon basis, then uniformly scaled up or down as needed.

PR #83, opened 2018-09-21
[SDPA-971] date picker

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-971

### Changed

1.  Added pikadate calendar picker.
2. Added fieldRpldatepicker - with ranged or single option.
3. Fixed warning on test regarding RplList component.

### Screenshots

screen shot 2018-09-21 at 5 11 05 pm
screen shot 2018-09-21 at 5 10 01 pm
screen shot 2018-09-21 at 5 09 44 pm


src/storybook-components/_data/demoData.js
```diff
@@ -932,7 +932,9 @@ const demoData = {
         tel: '',
         number: '',
         radio: null,
-        textArea: ''
+        textArea: '',
+        dateRange: ['', ''],
```
alan-cole: The resulting data is a string for a single field, and an array of 2 strings for the range field. Without further direction listed in the ticket, this may need to be reviewed to ensure it will work in the future required cases.

packages/Molecules/Form/formPikaday.vue
```diff
@@ -0,0 +1,303 @@
...
+    :disabled="disabled"
+    :placeholder="placeholder"
+    :readonly="readonly"
+    :name="inputName" />
+</template>
+
+<script>
+import { defaults } from 'lodash'
+import moment from 'moment'
+import Pikaday from 'pikaday'
```
alan-cole: While a pikaday widget does exist in the vue-form-generator library, it requires that pikaday is loaded from the head of the web page. If you check this line you can see how it expects window to have a pikaday variable loaded. This seems to go against how ripple is set up, as pikaday should be loaded into the scope of the form component. Because of this I have avoided using the existing element and created a custom wrapper formPikaday.vue.

PR #93, opened 2018-09-27
[SDPA-656] image gallery

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-656

### Changed

1.  Added Image Gallery

### Limitations
As this is using Vue Carousel - there are limitations:
* Loop (pressing back on the first item) causes carousel to flash right through all images to reach the last instead of showing the last item come in the from the left. This is a limitation of the library.
* Mobile does not show navigation in full view. This was not specified in design, and no swipe functionality is in AC. Users can close full screen and change view using the left/right arrows on the thumbnail view.
* As with the Card Carousel, there are issues with using any navigation other than the arrows due to limitations of the library.
* When changing breakpoint Vue Carousel will sometimes change the visible slide, this will cause the arrows to become out of sync with the shown card. When clicking on an arrow this may jump to a card 2 or more steps away. This is a limitation of the library.
* At some window sizes the thin grey border on the right side may be hidden from the Vue Carousel's overflow: hidden. This overflow hidden is needed for the carousel to properly swipe, and the hiding of the card's right border is likely caused by sub-pixel widths on the main column - the browser rounds up the single pixel border and draws it behind the hidden mask. A solution would probably need to look at avoiding sub-pixel widths in the grid.

### Screenshots

mobile-large-caption
mobile-large
mobile-thumb
desktop-large
desktop-thumb


packages/Atoms/Global/README.md
```diff
@@ -20,6 +20,7 @@ npm install @dpc-sdp/ripple-global --save
 ├── masonry-layout
 ├── moment
 ├── object-fit-images
+├── vue-carousel
```
tim-yao: Hey @alan-cole I think we should leave vue-carousel in card package and gallery package. Not need to be in the core.

packages/Atoms/Global/README.md
```diff
@@ -20,6 +20,7 @@ npm install @dpc-sdp/ripple-global --save
...
 ├── vue-focus
 └── vue-social-sharing
```
tim-yao: Same as social sharing, I don't think it should be in core.

PR #112, opened 2018-10-30
[SDPA-925] Updated search icon, text and input box alignment.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-925

### Changed

1.  Updated search icon, text and input box alignment.

### Screenshots

simulator screen shot - iphone x - 2018-10-30 at 12 58 33
simulator screen shot - iphone x - 2018-10-30 at 12 58 03

### DIFF
--- a/packages/Molecules/Search/SearchForm.vue
+++ b/packages/Molecules/Search/SearchForm.vue
@@ -133,6 +133,7 @@ export default {
     &__btn {
       background-color: transparent;
       border: 0;
+      padding: 0;
       width: $rpl-search-form-button-width;
       cursor: pointer;
 
@@ -145,6 +146,7 @@ export default {
       @include rpl_typography_ruleset($rpl-search-form-input-ruleset);
       background-color: transparent;
       border: 0;
+      padding: 1px;
       width: 100%;
       color: $rpl-search-form-input-text-color;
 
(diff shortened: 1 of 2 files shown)


packages/Molecules/Search/SearchForm.vue
```diff
@@ -145,6 +146,7 @@ export default {
       @include rpl_typography_ruleset($rpl-search-form-input-ruleset);
       background-color: transparent;
       border: 0;
+      padding: 1px;
```
alan-cole: This 1px is to match Chrome's default - it provides a little padding so when the focus indicator is on the field, the indicator isn't right up against the text.

packages/Organisms/SiteHeader/search.vue
```diff
@@ -76,6 +77,12 @@ export default {
       margin: $rpl-site-header-search-margin-xl;
     }
 
+    &__form {
+      &.rpl-search-form {
+        padding: 0;
```
alan-cole: This is to clear the default search form padding. It's not needed when the form is used in the header-search component.

tim-yao: Hi @alan-cole Just confirm with you, that change won't break any existing search forms like those on search result and event search page, right?

alan-cole: This should not affect those cases as the SCSS would compile down to:

```
.rpl-site-header-search__form.rpl-search-form {
  padding: 0;
}
```

which is adding a header-specific class. I have checked on the search results page (you can see it in the screenshot) and it has kept the search form component padding as needed.

PR #135, opened 2018-11-23
[SDPA-1319][SDPA-1302] adds document link component and zip icon.

dylankelly:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1319

### Changed

1.  Added RplDocumentLink component
1.  Added zip icon

### Screenshots
screen shot 2018-11-23 at 11 59 26 am


dylankelly:
Thinking about this I might move the computed logic to the nuxt side. This Ripple component should probably just be a presentation component. What do you think @tim-yao?

PR #136, opened 2018-11-23
VICPOLCOR-46: Add new ripple card for emergency contact, and card box…

komangariana:
… for its container.

**JIRA issue:** https://jira.ogilvy.com.au/browse/VICPOLCOR-46

### Changed

1.  Add new ripple card for emergency contact, and card box for its container

### Screenshots

card-box


packages/Molecules/Card/CardBox.vue
```diff
@@ -0,0 +1,87 @@
+<template>
+  <div class="rpl-card-box">
+    <div class="rpl-card-box-items">
+      <div class="rpl-card-box-item" v-for="(item, index) in cards" :key="index">
+        <component :is="item.name" v-bind="item.data" :trimFieldEventBus="isTrimmed(item.name) ? eventBus : null" :trimFieldUpdateOnResize="false"></component>
```
alan-cole: The eventBus is for invoking the trim update on components that use trimming. This can be useful in cases like the CardCarousel where additional libraries may affect the size of the container after it's mounted on the page. As the current component RplCardEmergencyContact does no use trimming, this is not needed right now.

In the case of future components being added, I still don't see any reason why invoking trimming would be needed in this case. I would consider removing this..

E.g.
```
<component :is="item.name" v-bind="item.data"></component>
```

packages/Molecules/Card/CardBox.vue
```diff
@@ -0,0 +1,87 @@
...
+      </div>
+    </div>
+  </div>
+</template>
+
+<script>
+import Vue from 'vue'
```
alan-cole: As with the above comment R.E eventBus - this could also be removed.

packages/Molecules/Card/CardBox.vue
```diff
@@ -0,0 +1,87 @@
...
+  name: 'RplCardBox',
+  props: {
+    cards: Array
+  },
+  components: {
+    RplCardEmergencyContact
+  },
+  data () {
+    return {
+      eventBus: new Vue(),
```
alan-cole: As with the above comment R.E eventBus - this and below could also be removed.

packages/Molecules/Card/CardBox.vue
```diff
@@ -0,0 +1,87 @@
...
+      // Define which cards use cardtrimfield. These will be given the eventBus
+      // object and their resize will be invoked on carousel size changes.
+      trimmedCards: [
+        'rpl-card-emergency-contact'
+      ]
+    }
+  },
+  methods: {
+    isTrimmed (name) {
```
alan-cole: As with the above comment R.E eventBus - this could also be removed.

packages/Molecules/Card/CardBox.vue
```diff
@@ -0,0 +1,87 @@
...
+  }
+}
+</script>
+
+<style lang="scss">
+@import "~@dpc-sdp/ripple-global/style";
+
+$rpl-card-box-padding: ($rpl-space * 15) 0 ($rpl-space * 9) !default;
+$rpl-card-box-margin: 0 (-$rpl-space-3) !default;
+$rpl-card-box-background-color: rgba(rpl-color('primary'), .15) !default;
```
tim-yao: I will suggest to make the Ripple components to fit the alpha theme, so maybe use our `light neutral` color as background here, and replace it with other colour in vicpol project.

komangariana:
Hi @alan-cole @tim-yao , 

I've updated the PR based on your feedback. 
Kindly review, 

Thank you,

PR #139, opened 2018-11-23
[SDPA-1077] Updated breadcrumb to sit closer to header.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1077

### Changed

1.  Updated breadcrumb to sit closer to header.

Note on implementation:
This is the least disruptive implementation to reduce breadcrumb spacing.
If breadcrumb does not display, page will render as it usually does. If breadcrumb does display, then all items below breadcrumb will sit slightly higher than they normally would.

The alternative way is to reduce the padding on `.rpl-above-content__inner`. If we do that then all items will sit slightly higher, whether breadcrumbs display or not. This means tickets like SDPA-986 will need to be redone as the logo will once again overlap the image.

I am happy to implement either way, but have chosen this way first as it's the simplest. Please let me know if you would like an alternative implementation.

### Screenshots

screen shot 2018-11-23 at 5 04 19 pm

### DIFF
--- a/packages/Molecules/Breadcrumbs/index.vue
+++ b/packages/Molecules/Breadcrumbs/index.vue
@@ -41,7 +41,7 @@ export default {
   $rpl-breadcrumbs-text-color: rpl-color('extra_dark_neutral') !default;
   $rpl-breadcrumbs-separator-color: rpl-color('dark_neutral') !default;
   $rpl-breadcrumbs-separator-padding: 0 ($rpl-space-2) 0 ($rpl-space) !default;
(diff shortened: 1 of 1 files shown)


tim-yao:
Hi @alan-cole as dissucced, I prefer to change spacing on `.rpl-above-content__inner`. And we need to remove the margin on breadcrumbs item as well.

alan-cole:
Made changes on https://github.com/dpc-sdp/ripple/pull/145

PR #149, opened 2018-11-30
VICPOLCOR-135: Fix issue - expanded image not displayed correctly.

komangariana:
**JIRA issue:** https://jira.ogilvy.com.au/browse/VICPOLCOR-135

### Changed

Add event on page change into carousel, and set prop navTo accordingly.
This will fix expanded image issue on mobile when user swipe the carousel.

### Screenshots

https://screencast.com/t/8tkexoHZ

### DIFF
--- a/packages/Organisms/ImageGallery/ImageGallery.vue
+++ b/packages/Organisms/ImageGallery/ImageGallery.vue
@@ -36,6 +36,7 @@
           :loop="true"
           :mouseDrag="false"
           :paginationEnabled="false"
+          @pageChange="onPageChange"
         >
           <slide v-for="(item, index) in galleryData" :key="index" class="rpl-image-gallery__large">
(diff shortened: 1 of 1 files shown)


packages/Organisms/ImageGallery/ImageGallery.vue
```diff
@@ -9,6 +9,7 @@
       :loop="true"
       :mouseDrag="false"
       :paginationEnabled="false"
+      @pageChange="onPageChange"
```
alan-cole: This event `@pageChange="onPageChange"` will need to be added to the modal carousel as well (see line 38) in order for both fullscreen and small screen carousels to be in sync.

komangariana: Hi @alan-cole @tim-yao , 

I've updated the PR. 
Kindly review, 

Thank you.

PR #183, opened 2019-01-07
[SDPA-1477] Adds video variant to rpl-campaign-secondary.

dylankelly:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1477

### Changed

1.  Adds ability to add video to RplCampaignSecondary.
2. Change mediaLink prop type on RplEmbeddedVideo from String to Object

### Screenshots
screen shot 2019-01-07 at 11 44 48 am

### DIFF
--- a/packages/Organisms/CampaignSecondary/index.vue
+++ b/packages/Organisms/CampaignSecondary/index.vue
@@ -1,9 +1,10 @@
 <template>
   <div class="rpl-campaign-secondary">
     <div class="rpl-campaign-secondary__row">
-      <div class="rpl-campaign-secondary__left">
+      <div v-if="image || video" class="rpl-campaign-secondary__left">
         <div class="rpl-campaign-secondary__content">
           <img v-if="image" :src="image" alt="" class="rpl-campaign-secondary__image" />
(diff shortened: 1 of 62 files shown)


packages/Organisms/CampaignSecondary/index.vue
```diff
@@ -116,6 +120,10 @@ export default {
         padding: $rpl-campaign-secondary-content-padding-s;
       }
     }
+    &__video {
+      .rpl-embedded-video-iframe-container iframe {border: 0}
```
tim-yao: Hey @dylankelly this reminds me the sass lint is not working in vue template file. If it works, it will throw warning I think. Suggested way is:
```
.rpl-embedded-video-iframe-container {
  iframe {
    border: 0;
  }
}
```
Another thing I think is this css may should be added to `RplEmbeddedVideo` component which shouldn't have a border, right?

dylankelly: ah, from your solution direction it sounded like you just wanted this on the campaign secondary.  Have updated.

PR #188, opened 2019-01-08
[SDPA-1205] Reset pagination arrow on totalStep change.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1205

### Changed

1. Reset pagination arrow on totalStep change.
2. Update Search Page to allow for testing.
3. Remove unused property from Search Results story.

### Screenshots

screen shot 2019-01-08 at 5 41 38 pm

### DIFF
--- a/src/storybook-components/SearchPage.vue
+++ b/src/storybook-components/SearchPage.vue
@@ -34,10 +34,10 @@
       <rpl-row row-gutter class="demo-main">
         <rpl-col cols="full" :colsBp="defaultCols">
           <rpl-search-results
-            :searchResults="noResults ? [] : [mock.searchResult, mock.searchResult]"
-            :pager="noResults ? undefined : mock.pagination"
+            :searchResults="subsetResults"
+            :pager="pagerOptions"
             :responseSize="noResults ? 0 : mock.searchResults.responseSize"
-            :count="noResults ? 0 : mock.searchResults.count"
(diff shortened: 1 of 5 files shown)


src/storybook-components/SearchPage.vue
```diff
@@ -118,13 +121,60 @@ export default {
...
+      } else {
+        let returnedResults = []
+        const searchResultJSON = JSON.stringify(this.mock.searchResult)
+        for (let i = 0; i < this.totalResultCount; i++) {
+          let newResult = JSON.parse(searchResultJSON)
+          newResult.title = `${this.searchFor} result ${i + 1}`
+          returnedResults.push(newResult)
+        }
+        return returnedResults
+      }
```
tim-yao: Hi @alan-cole I am confused for this part, are they not need to be inside our component? Does all the place we use search results need those logic?

alan-cole: In order to ensure the issue has been fixed in ripple, we need a way to test that the search results and pagination functionality work. This means we need a way to simulate a search and changing between pages.

Before there was no such functionality in ripple - it would show static results that couldn't be altered and clicking pagination would show an alert, but not change page.

This logic is here for simulation but is not useful in a real world scenario. Nuxt uses different implementation to display real results.

PR #195, opened 2019-01-15
[SDPA-1622] VFG RplMarkup field added.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1622

### Changed

1.  VFG RplMarkup field added.

### Screenshots

screen shot 2019-01-16 at 8 41 33 am

### DIFF
--- a/packages/Molecules/Form/fields/fieldRplmarkup.vue
+++ b/packages/Molecules/Form/fields/fieldRplmarkup.vue
@@ -1,0 +1,52 @@
+<template>
+  <div class="rpl-markup" v-html="schema.markup"></div>
+</template>
+
+<script>
+import { abstractField } from 'vue-form-generator'
+
+export default {
+  mixins: [abstractField]
+}
+</script>
+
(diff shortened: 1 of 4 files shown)


packages/Molecules/Form/fields/fieldRplmarkup.vue
```diff
@@ -0,0 +1,52 @@
...
+<style lang="scss">
+@import "~@dpc-sdp/ripple-global/scss/settings";
+@import "~@dpc-sdp/ripple-global/scss/tools";
+
+$tide-link-color: rpl-color('primary') !default;
+
+.rpl-markup {
+  @include rpl_typography;
+
+  // The below should be cleaned up. It's taken from TideWysiwyg component.
```
alan-cole: As the WYSIWYG component exists in Nuxt, there is a duplication of styles. Ideally we would use the same component in both cases, but will require some effort and planning to move WYSIWYG back into Ripple.

PR #200, opened 2019-01-18
[SDPA-1729] Adds wrapper for vueMultiselect to use with object arrays.

dylankelly:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1729

### Changed

1.  Adds fieldRplSelect VFG component which wraps vuemultiselect component to handle arrays of objects. Currently only supports single selection.
1.  Fixes required indicator propagating to child labels

### DIFF
--- a/packages/Molecules/Form/index.vue
+++ b/packages/Molecules/Form/index.vue
@@ -16,8 +16,9 @@
 <script>
 import Vue from 'vue'
 import RplFormAlert from './formAlert'
-import VueFormGenerator from 'vue-form-generator'
 import Multiselect from 'vue-multiselect'
+import VueFormGenerator from 'vue-form-generator'
+import fieldRplselect from './fields/fieldRplselect.vue'
 import fieldRplslider from './fields/fieldRplslider.vue'
 import fieldRplcheckbox from './fields/fieldRplcheckbox.vue'
 import fieldRplchecklist from './fields/fieldRplchecklist.vue'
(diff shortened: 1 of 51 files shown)


src/storybook-components/_data/demoData.js
```diff
@@ -1323,7 +1323,8 @@ const demoData = {
         dateRange: ['', ''],
         date: '',
         checkbox: true,
-        vuemultiselect: null,
+        select: [],
```
alan-cole: Is this new select model used anywhere?

packages/Molecules/Form/index.vue
```diff
@@ -173,7 +173,10 @@ export default {
 
   .error {
     input,
-    textarea {
+    textarea,
+    .multiselect__tags,
+    .form-control
+    {
```
alan-cole: If the SCSS linting worked in .vue files, I think this one would be picked up. { should be on the same line as the last rule.

packages/Molecules/Form/stories.js
```diff
@@ -11,7 +11,7 @@ storiesOf('Molecules/Form', module)
   .addDecorator(VueInfoAddon)
   .add('Form', withReadme(readme, () => ({
     components: { RplForm },
-    template: `<rpl-form :formData="formData"></rpl-form>`,
+    template: `<rpl-form :formData="formData" :submitHandler="submitHandler" :is-new-model="isNewModel" :options="options"></rpl-form>`,
```
alan-cole: Should there be some consistency between prop naming, should `:is-new-model` be `:isNewModel`?

packages/Molecules/Form/index.vue
dylankelly: thanks, need to fix that linter!

PR #205, opened 2019-01-22
v1.0.0-alpha.45

dylankelly:

### Changed

1.

### DIFF
--- a/packages/Organisms/Markup/package.json
+++ b/packages/Organisms/Markup/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dpc-sdp/ripple-markup",
-  "version": "1.0.0-alpha.43",
+  "version": "1.0.0-alpha.45",
   "description": "A component to wrap wysiwyg fields and process markup.",
   "license": "Apache-2.0",
   "main": "index.vue",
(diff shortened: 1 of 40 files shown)


packages/Organisms/Markup/package.json
```diff
@@ -10,9 +10,9 @@
     "access": "public"
   },
   "dependencies": {
-    "@dpc-sdp/ripple-global": "1.0.0-alpha.43",
-    "@dpc-sdp/ripple-document-link": "1.0.0-alpha.43",
     "@dpc-sdp/ripple-button": "1.0.0-alpha.43",
+    "@dpc-sdp/ripple-document-link": "1.0.0-alpha.43",
```
tim-yao: need to be `1.0.0-alpha.45`

dylankelly: ah good catch, i wonder why that didn't get updated by lerna?

tim-yao: @dylankelly Lerna will increment all package's current version to next version, but when you rebase/merge branch your working branch code is outdated, the version is not same as the current version, so it didn't get picked by Lerna. It happened before, so I did check it. 🤓

PR #218, opened 2019-01-29
[SDPA-1789] Adds vue-on-click-outside to close rplchecklist on click outside

dylankelly:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1789

### Changed

1.  Adds vue-on-click-outside to close rplchecklist on click outside


dylankelly:
Note - This is a different package then the previously used one and is not a directive. In testing it doesn't display the same errors as that package.

PR #222, opened 2019-01-30
[SDPA-1667] Add HonourRoll card. Update Search results.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1667

### Changed

1.  Add HonourRoll card. Update Search results.

### Screenshots

screen shot 2019-01-30 at 15 19 29


package-lock.json
```diff
@@ -3437,9 +3437,9 @@
       "dev": true
```
tim-yao: If there is no package update, I will suggest not commit package-lock file, to reduce the conflicts.

PR #224, opened 2019-01-31
[SDPA-1769] Clear search form button

dylankelly:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1769

### Changed

1.  Adds rplClearsearch field to clear form data
1.  Refactors rplCheckbox to use computed value instead of watchers
1.  Adds watchers to RplSelect and RplDatepicker to reset on null value 

Note: Does not implement inline formatting with submit button. I feel like this should be part of the implementation of the form to keep this field reusable.

### Screenshots
screen shot 2019-01-31 at 11 20 43 am

### DIFF
--- a/packages/Molecules/Form/Checkbox.vue
+++ b/packages/Molecules/Form/Checkbox.vue
@@ -39,7 +39,6 @@ export default {
   },
   data () {
     return {
-      checked: this.value,
       id: null
     }
   },
@@ -47,8 +46,8 @@ export default {
     this.id = this._uid
   },
   methods: {
-    updateValue () {
-      this.$emit('input', this.checked)
+    updateValue (val) {
+      this.$emit('input', val)
     },
     labelClick () {
       this.$refs['input_checkbox'].focus()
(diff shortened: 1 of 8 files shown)


packages/Molecules/Form/Checkbox.vue
```diff
@@ -57,13 +56,20 @@ export default {
...
-      this.updateValue()
-      this.updateChange(val)
+  computed: {
+    checked: {
+      set (val) {
+        this.updateValue(val)
+        this.updateChange(val)
+      },
+      get () {
+        if (this.value === null || this.value === undefined) {
```
tim-yao: Hi @dylankelly I am just wondering, instead of adding this `null` condition in watch for all form field, should we move this logic to the place set to `null` https://github.com/dpc-sdp/ripple/pull/224/files#diff-523d7114849b5c86e4a722370037e520R18 ? What about those Vue form generator fields? Are they all working without add this logic?

dylankelly: the problem here is that each field (checkbox, date, etc) has a different default (initial blank value). Ie: checkbox is false, datepicker range is ['', '']. I feel like this is better placed in the fields themselves, otherwise RplClearsearch needs to know about the internal logic of all these different fields. Also this is not just for clear search, it allows you to initialise each field with null and it be handled properly. The default fields handle null as a blank value already.

packages/Molecules/Form/fields/fieldRplclearform.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+  <button class="rpl-clearform" @click="clearForm">
+    <rpl-text-icon :text="schema.buttonText" symbol="cross_circle" placement="before" color="danger" size="m" />
```
alan-cole: An alternative to setting size to 'm' would be to set size="0.8333". By doing this we can remove max-width and max-height values from the SCSS below. I'm not too fussed either way, but this is how it has been done in other cases.

PR #226, opened 2019-01-31
[SDPA-1762] Adds RplFieldset component, adds grouped fields stories.

dylankelly:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1762

### Changed

1.  Adds RplFieldset component to encapsulate fieldset wrapper and grid.
1. Adds example story for using grouped fields and nested object for address component (see form story)

### Screenshots
screen shot 2019-01-31 at 5 12 47 pm

### DIFF
--- a/packages/Molecules/Form/Fieldset.vue
+++ b/packages/Molecules/Form/Fieldset.vue
@@ -1,0 +1,56 @@
+<template>
+  <fieldset class="rpl-form-group">
+    <div class="rpl-form-group__inner">
+      <slot></slot>
+    </div>
+  </fieldset>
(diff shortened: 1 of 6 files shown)


dylankelly:
@tim-yao - The documentation for using form groups is within the forms story, best to lead by example.  I think the best place for implementation documentation will be on the nuxt side as this is where there will need to be changes (see https://github.com/dpc-sdp/vic-gov-au/pull/450)

We should consider a way to add custom text to the Readme, as is anything added will be removed after the docs are regenerated.

PR #258, opened 2019-02-19
[SDPA-1759] Fixed doc link download name issue

tim-yao:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1759

### Changed

1.  Changed download attr value to empty string instead of true, so file name won't be rename.


tim-yao:
As it's fix for UAT fail, will let it QA in Nuxt. I have done a QA in storybook and it passed.

PR #264, opened 2019-02-20
[SDPA-1976] Card details centred

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1976

### Changed

1.  Added 'center' prop to Card Content, and applied in Honour card (when no image).

### Screenshots

screen shot 2019-02-20 at 16 13 58-fullpage

### DIFF
--- a/packages/Molecules/Card/stories.js
+++ b/packages/Molecules/Card/stories.js
@@ -105,7 +105,7 @@ storiesOf('Molecules/Card', module)
   })))
   .add('Card Content (base)', withReadme(readmeCardContent, () => ({
     components: { RplCardContent },
-    template: `<rpl-card-content :link="link" :image="image" :border="border" :type="type">
+    template: `<rpl-card-content :link="link" :image="image" :border="border" :type="type" :center="center">
   <div v-html="content"></div>
 </rpl-card-content>`,
     data () {
--- a/src/storybook-components/PageCards.vue
+++ b/src/storybook-components/PageCards.vue
@@ -1,0 +1,112 @@
+<template>
+  <rpl-base-layout class="demo">
+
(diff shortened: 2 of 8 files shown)


packages/Molecules/Card/stories.js
```diff
@@ -105,7 +105,7 @@ storiesOf('Molecules/Card', module)
   })))
   .add('Card Content (base)', withReadme(readmeCardContent, () => ({
     components: { RplCardContent },
-    template: `<rpl-card-content :link="link" :image="image" :border="border" :type="type">
+    template: `<rpl-card-content :link="link" :image="image" :border="border" :type="type" :center="center">
   <div v-html="content"></div>
```
tim-yao: I feel this should be a `slot` in stead of `prop`, how do you think? @alan-cole ?
But we can ignore it in this PR as it's a big change.

src/storybook-components/PageCards.vue
```diff
@@ -0,0 +1,112 @@
+<template>
+  <rpl-base-layout class="demo">
```
tim-yao: @alan-cole I think we have a landing page testing all type of cards in page layout already, will this become a duplication?
Instead of creating a new demo page for cards, maybe add a story for all cards inside card story?

alan-cole: This was added more for QA - to be able to compare how cards look. I wasn't able to find a screen that really showed this elsewhere (closest was landing page). Happy to remove if you'd prefer.

packages/Molecules/Card/stories.js
alan-cole: I think we'd need further discussion as to how slots would work for this.

PR #269, opened 2019-02-22
[SDPA-2054] Adds option to disable entity decoding an disables by def…

dylankelly:
…ault.

**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2054

### Changed

1.  Adds option to disable entitiyDecoding on RplMarkup, sets to false by default.

### Screenshots
screen shot 2019-02-22 at 2 55 15 pm

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -11,7 +11,7 @@ import RplGlobal from './../packages/Atoms/Global'
 import RplMarkupExamplePlugins from './../packages/Organisms/Markup/examplePlugins'
 
 // Install Ripple Global plugin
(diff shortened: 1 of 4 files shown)


dylankelly:
@tim-yao I did a bit more research on this, it appears to be a bug with Cheerio. https://github.com/cheeriojs/cheerio/issues/627, Seems to describe the issue with apostrophes. The issue is closed but only because it is a duplicate of the open issue https://github.com/cheeriojs/dom-serializer/issues/1. As said here https://github.com/cheeriojs/cheerio/issues/627#issuecomment-69839794 the workaround is to set decodeEntities to false which this PR implements.

PR #275, opened 2019-02-26
[SDPA-1870] Fill select fields on load if the model already has a value.

dylankelly:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1870

### Changed

1.  Makes fieldRplSelect set default value if it has been set in the model.
1.  Fix height on select field.
1.  Moved mounted hook in fieldRplchecklist to top for better readability.

### DIFF
--- a/packages/Molecules/Form/fields/fieldRplchecklist.vue
+++ b/packages/Molecules/Form/fields/fieldRplchecklist.vue
@@ -62,6 +62,12 @@ export default {
       labelText: ''
     }
   },
+  mounted () {
+    if (this.value) {
+      this.updateSize()
+      this.setCheckedValues()
+    }
+  },
   watch: {
     value (newVal, oldVal) {
       this.updateSize()
--- a/packages/Molecules/Form/fields/fieldRplchecklist.vue
+++ b/packages/Molecules/Form/fields/fieldRplchecklist.vue
@@ -63,10 +63,11 @@ export default {
     }
   },
   mounted () {
-    if (this.value) {
+    if (typeof window !== 'undefined') {
+      window.addEventListener('resize', this.updateSize)
       this.updateSize()
-      this.setCheckedValues()
     }
+    this.setCheckedValues()
   },
   watch: {
     value (newVal, oldVal) {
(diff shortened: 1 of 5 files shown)


anthony-malkoun:
Hey @dylankelly is there any reason this is against `master` and not `develop`?

src/storybook-components/_data/demoData.js
```diff
@@ -1416,12 +1416,14 @@ const demoData = {
...
-        address: {},
+        address: {
+          state: 'vic'
+        },
         checkbox: true,
         select: '',
         multiselect: null,
-        checklistlistbox: null,
-        checklistdropdown: null,
+        checklistlistbox: ['topic_a'],
```
alan-cole: It's good to test how a form element responds when it has a default value. but there's also benefit testing an item when it isn't set with a value. I'm not sure how best to provide all potential states - would including these fields as null alongside these prepopulated ones be better? It's a pass, but maybe a broader discussion for the future.

tim-yao:
> Hey @dylankelly is there any reason this is against `master` and not `develop`?

Hi @anthony-malkoun This Ripple UI project has no `develop` branch at this stage, same as all tide modules. e.g `https://github.com/dpc-sdp/tide_search`. The reason is there is no develop or production concept at what the work flow in our team for this part. They are been used more like dev latest version at this stage, happy to discuss if it's confusing. 😺

src/storybook-components/_data/demoData.js
tim-yao: Hi @alan-cole We will add Vue test utils soon https://vue-test-utils.vuejs.org/ for better component unit testing. Demo data will mostly be used for visual demos.

packages/Molecules/Form/fields/fieldRplchecklist.vue
```diff
@@ -62,6 +62,13 @@ export default {
       labelText: ''
     }
   },
+  mounted () {
+    if (typeof window !== 'undefined') {
+      window.addEventListener('resize', this.updateSize)
+      this.updateSize()
+    }
+    this.setCheckedValues()
+  },
```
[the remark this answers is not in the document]
dylankelly: I moved this because I didn't see it all the way down the bottom and ended up rewriting it. We should probably add linter rule for ordering instance methods. I'd recommend we follow vue style guide for this https://vuejs.org/v2/style-guide/#Component-instance-options-order-recommended.

src/storybook-components/_data/demoData.js
dylankelly: @alan-cole in this case we need the default value to show functionality that was missing (selects weren't initialising with default value).

PR #283, opened 2019-03-01
[SDPA-2131]Added default variables for rpl-markup and underlines in rpl-textlink

FleetAdmiralButter:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPSUP-124

### Changed

1.  Added default variable for rpl-markup
2. Added default variables for rpl-text-link


packages/Organisms/Markup/Markup.vue
```diff
@@ -42,24 +42,26 @@ export default {
 </script>
 
 <style lang="scss">
-@import "~@dpc-sdp/ripple-global/style";
+  @import "~@dpc-sdp/ripple-global/style";
```
tim-yao: Shouldn't have indent here and below

PR #289, opened 2019-03-12
[SDPA-2187] Implement an interface for defining child column breakpoints

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2187

### Changed

1.  Implement an interface for defining child column breakpoints

### Screenshots

Screen Shot 2019-03-12 at 5 33 51 pm


packages/Atoms/Global/mixins/ProvideChildCols.js
```diff
@@ -0,0 +1,7 @@
+const ProvideChildCols = {
```
tim-yao: No capital for first letter `provideChildCols`

PR #292, opened 2019-03-17
[SDPA-2221] Re implements Rpl-select with focus on accessibility

dylankelly:
Adds presentational prop to checkbox and adds to multselect.
Add aria labelled by for button
Only render native select on mobile.

**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2221

### Changed

1.  Created Rpl-Select component based upon https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html spec


packages/Molecules/Form/package.json
```diff
@@ -16,6 +16,7 @@
     "vue-form-generator": "^2.2.2",
     "vue-multiselect": "^2.1.3",
     "vue-on-click-outside": "^1.0.3",
-    "vue-scrollto": "^2.13.0"
+    "vue-scrollto": "^2.13.0",
+    "vue-media": "^1.1.1"
```
tim-yao: @dylankelly Are you using `vue-media` any place? Just couldn't find it. Also we have a `breakpoint` mixins https://github.com/dpc-sdp/ripple/blob/master/packages/Atoms/Global/mixins/breakpoint.js which you may can use for same purpose.

src/storybook-components/_data/demoData.js
```diff
@@ -1591,40 +1591,14 @@ const demoData = {
             placeholder: 'Select multiple topics',
             values: [{value: 'topic_a', name: 'Topic A'}, {value: 'topic_b', name: 'Topic B'}, {value: 'topic_c', name: 'Topic C'}, {value: 'topic_d', name: 'Topic D'}]
           },
-
-          {
-            type: 'rplchecklist',
```
tim-yao: This will introduce a breaking change in that version, we may keep it as deprecated so we won't have trouble to update all sites.
https://github.com/search?q=org%3Adpc-sdp+rplchecklist&type=Code

We may need to add a console warning in our Ripple form when `rplchecklist` is used. https://css-tricks.com/approaches-to-deprecating-code-in-javascript/

src/storybook-components/_data/demoData.js
```diff
@@ -1591,40 +1591,14 @@ const demoData = {
...
             placeholder: 'Select a single topic',
-            selectOptions: {
-              trackBy: 'id',
-              label: 'name',
-              closeOnSelect: true,
-              searchable: false,
-              showLabels: false
-            },
-            values: [{id: 'topic_a', name: 'Topic A'}, {id: 'topic_b', name: 'Topic B'}, {id: 'topic_c', name: 'Topic C'}, {id: 'topic_d', name: 'Topic D'}]
+            values: [{id: 'topic_a', name: 'Topic A'}, {id: 'topic_b', name: 'Topic B'}, {id: 'topic_c', name: 'Topic C'}, {id: 'topic_d', name: 'Topic D'}, {id: 'topic_e', name: 'Topic e'}, {id: 'topic_f', name: 'Topic f'}, {id: 'topic_g', name: 'Topic g'}, {id: 'topic_h', name: 'Topic h'}]
```
tim-yao: This `rplselect` changes also will bring breaking changes, https://github.com/search?q=org%3Adpc-sdp+rplselect&type=Code
We may can provide a new one to deprecate the old one, something like `rplselectV2`.

packages/Molecules/Form/package.json
dylankelly: ah cool, i hadn't seen that mixin.

src/storybook-components/_data/demoData.js
dylankelly: I dont think we will need to do that, because the external interface is (relatively) the same. Implementers will just need to remove `selectOptions` key, which won't do anything anyway. We could support using `selectOptions.multiple` as well as `multiselect` if it was used alot, but as this was only recently added, I can't see it being used widely.

If vue-multiselect is needed, its available as part of vue-form-generator.

src/storybook-components/_data/demoData.js
dylankelly: This PR doesn't remove rplChecklist (it removes the story though, which I think we do want to do now so others dont use it). Good point about the deprecation message, will probably need to add to RplForm

packages/Molecules/Form/fields/fieldRplselect.vue
```diff
@@ -1,118 +1,384 @@
...
-    :max="schema.max || null"
-    :disabled="disabled"
-    @input="updateSelected"
-    @search-change="onSearchChange"
-    @tag="addTag"
-    :option-height="selectOptions.optionHeight"
-  >
-  </multiselect>
+  <div class="rpl-select" :class="{'rpl-select--open' : isOpen}">
+    <div v-if="$breakpoint.s" class="rpl-select__native">
```
alan-cole: Currently we have this field visually hidden at breakpoint "s" and above. I'm finding that there's now 2 ways to control the select box:
* by focusing on the `<select>` option below (hidden by screen but still works for screen readers)
OR
* by focusing on the rpl-select__trigger `<div>` below, which works for both screen readers and visually.

I think there should only be one input method to avoid confusion.

Also - using `$breakpoint.s` means any screen size below `s` (576 and under) will be false, meaning this will NOT show on mobile.

packages/Molecules/Form/fields/fieldRplselect.vue
```diff
@@ -1,118 +1,384 @@
...
+        document.removeEventListener('click', this.testOutside)
       }
     },
-    onSearchChange (searchQuery, id) {
-      let onSearch = this.selectOptions.onSearch
-      if (typeof onSearch === 'function') {
-        onSearch(searchQuery, id, this.options)
+    preventKeyboardScroll (e) {
+      // up down and space keys
+      const keys = ['Up', 'ArrowUp', 'Down', 'ArrowDown', ' ', 'Space']
```
alan-cole: Trailing space is throwing an error.

packages/Molecules/Form/fields/fieldRplchecklist.vue
```diff
@@ -196,6 +196,37 @@ export default {
 
       this.labelText = str
     },
+    updateSize () {
```
alan-cole: This whole function is a duplicate.

packages/Molecules/Form/fields/fieldRplselect.vue
```diff
@@ -1,118 +1,384 @@
...
+        class="rpl-select__trigger"
+        tabindex="0"
+        aria-haspopup="listbox"
+        :aria-labelledby="`${schema.model}-rpl-select-trigger ${schema.model}-rpl-select-value`"
+        :aria-expanded="isOpen ? 'true' : false"
+        role="button"
+        ref="trigger"
+        @click="toggleOpen"
+        v-on:keyup.enter.prevent="toggleOpen"
+      >
```
alan-cole: Using NVDA on windows 10 with IE11 and Edge, I found that IE11 wouldn't read out the single select options as you press arrow up and down (so a user wouldn't know what to select), and on Edge, single select DID work, but multiselect did not.

packages/Molecules/Form/fields/fieldRplselect.vue
```diff
@@ -1,118 +1,384 @@
...
+    padding: $rpl-space-3 $rpl-form-element-padding-m-horizontal;
+    background: $rpl-form-element-bg-color;
+
+    &:nth-child(odd) {
+      background-color: rpl-color("white");
+    }
+
+    &:hover {
+      background-color: rpl-color("secondary");
+      color: white;
```
alan-cole: `rpl-color("white");` as an alternative.

packages/Molecules/Form/fields/fieldRplselect.vue
```diff
@@ -1,118 +1,384 @@
...
+    }
+
+    &--selected[aria-selected="true"] {
+      background-color: rpl-color("primary");
+      color: white;
```
alan-cole: `rpl-color("white");` as an alternative.

packages/Molecules/Form/fields/fieldRplselect.vue
```diff
@@ -1,118 +1,384 @@
...
+    }
+
+    &--focussed {
+      color: white !important;
```
alan-cole: `rpl-color("white");` as an alternative.

packages/Molecules/Form/fields/fieldRplselect.vue
```diff
@@ -1,118 +1,384 @@
...
+  &__listitem {
+    padding: $rpl-space-3 $rpl-form-element-padding-m-horizontal;
+    background: $rpl-form-element-bg-color;
+
+    &:nth-child(odd) {
+      background-color: rpl-color("white");
+    }
+
+    &:hover {
+      background-color: rpl-color("secondary");
```
alan-cole: Especially with colours, these may be better served as $variables for possible theme changes.

packages/Molecules/Form/fields/fieldRplselect.vue
```diff
@@ -1,118 +1,384 @@
...
-    :disabled="disabled"
-    @input="updateSelected"
-    @search-change="onSearchChange"
-    @tag="addTag"
-    :option-height="selectOptions.optionHeight"
-  >
-  </multiselect>
+  <div class="rpl-select" :class="{'rpl-select--open' : isOpen}">
+    <div v-if="$breakpoint.s" class="rpl-select__native">
+      <select :multiple="schema.multiselect" v-model="value">
```
alan-cole: For mobile, the use of a label here (possibly part of the schema object?) would help with mobile screen readers.

packages/Molecules/Form/fields/fieldRplselect.vue
```diff
@@ -1,118 +1,384 @@
...
+        role="button"
+        ref="trigger"
+        @click="toggleOpen"
+        v-on:keyup.enter.prevent="toggleOpen"
+      >
+        <template v-if="value && value.length > 0">
+          <span :id="`${schema.model}-rpl-select-value`">{{selectedTitles}}</span>
+        </template>
+        <span v-else :id="`${schema.model}-rpl-select-trigger`">{{placeholder}}</span>
+        <rpl-icon class="rpl-select__trigger-icon" symbol="down" color="primary" />
```
alan-cole: While showing this icon for multi-select on mobile makes sense. It looks strange on desktop with a small window, having an arrow beside a multi-select `<select>` box.

Screen Shot 2019-03-27 at 6 55 06 pm

Not sure what to do about that.

packages/Molecules/Form/fields/fieldRplselect.vue
dylankelly: yes, i think this was around the wrong way when you looked at it.

packages/Molecules/Form/fields/fieldRplselect.vue
dylankelly: fixed that

packages/Molecules/Form/fields/fieldRplchecklist.vue
dylankelly: merge problem, fixed

packages/Molecules/Form/fields/fieldRplselect.vue
dylankelly: Have added focus and selected colours as variables

packages/Molecules/Form/fields/fieldRplselect.vue
dylankelly: have added id on select element so it will bel labelled by the VFG generated label.

packages/Molecules/Form/fields/fieldRplselect.vue
dylankelly: agreed, the icon is weird for multiselect on mobile, have removed.

packages/Molecules/Form/fields/fieldRplselect.vue
dylankelly: wasn't managing focus, in my testing it now reads options in both NVDA and voiceover

PR #307, opened 2019-03-22
[SDPA-1604] Adds better screen reader support for pagination.

dylankelly:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1604

### Changed

1.  Give pagination navigation role
1.  Set aria-current="page" on active item
1.  Move slash from pseudo element to actual span and add aria-hidden attribute to prevent screen readers from announcing it.
1. Use button with disabled attribute for active number (and give a label that screen readers can use)

### DIFF
--- a/packages/Molecules/Pagination/index.vue
+++ b/packages/Molecules/Pagination/index.vue
@@ -1,9 +1,17 @@
 <template>
-  <div class="rpl-pagination">
+  <nav class="rpl-pagination" role="navigation" :aria-label="label || 'search'">
     <ol class="rpl-pagination__list">
-      <li v-for="n in visibleStepRange" :key="n" class="rpl-pagination__list-item">
-        <button v-if="n !== currentStep" @click="gotoStep(n)" class="rpl-pagination__step"><span>Go to step </span>{{ n }}</button>
-        <span v-else class="rpl-pagination__step-current">{{ n }}</span>
+      <li v-for="n in visibleStepRange" :aria-current="n === currentStep ? 'page' : false" :aria-disabled="n === currentStep ? 'true' : false" :key="n" class="rpl-pagination__list-item">
+        <button
+          @click="gotoStep(n)"
+          class="rpl-pagination__step"
+          :class="{'rpl-pagination__step-current' : n === currentStep}"
+          :disabled="n === currentStep"
+          >
+          <span class="rpl-pagination__step-label">{{n !== currentStep ? 'Go to step' : 'current page'}}</span>
+          {{ n }}
+        </button>
+        <span aria-hidden="true" v-if="n < visibleStepRange" class="rpl-pagination__list-item-slash">/</span>
       </li>
     </ol>
     <div class="rpl-pagination__controls">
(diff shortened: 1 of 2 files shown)


packages/Molecules/Pagination/index.vue
```diff
@@ -1,9 +1,17 @@
 <template>
-  <div class="rpl-pagination">
+  <nav class="rpl-pagination" role="navigation" :aria-label="label || 'search'">
```
alan-cole: Role of Navigation is implied by the element `<nav>`.
https://www.w3.org/TR/html5/sections.html#the-nav-element
recommended not to set:
> Allowed ARIA role attribute values:
>     navigation role (default - do not set). 
> In the majority of cases setting an ARIA role and/or aria-* attribute that matches the default implicit ARIA semantics is unnecessary and not recommended as these properties are already set by the browser.

packages/Molecules/Pagination/index.vue
```diff
@@ -25,6 +33,7 @@ import RplIcon from '@dpc-sdp/ripple-icon'
 export default {
   name: 'RplPagination',
   props: {
+    label: String,
```
alan-cole: Would setting a default of 'search' here instead of in the template be better?

packages/Molecules/Pagination/index.vue
```diff
@@ -1,9 +1,17 @@
...
-        <button v-if="n !== currentStep" @click="gotoStep(n)" class="rpl-pagination__step"><span>Go to step </span>{{ n }}</button>
-        <span v-else class="rpl-pagination__step-current">{{ n }}</span>
+      <li v-for="n in visibleStepRange" :aria-current="n === currentStep ? 'page' : false" :aria-disabled="n === currentStep ? 'true' : false" :key="n" class="rpl-pagination__list-item">
+        <button
+          @click="gotoStep(n)"
+          class="rpl-pagination__step"
+          :class="{'rpl-pagination__step-current' : n === currentStep}"
+          :disabled="n === currentStep"
+          >
+          <span class="rpl-pagination__step-label">{{n !== currentStep ? 'Go to step' : 'current page'}}</span>
```
alan-cole: I think the AC requested 'Go to step' be read as 'Go to page'.

packages/Molecules/Pagination/index.vue
dylankelly: Ah this is one of those ideal world vs real world things - https://www.w3.org/WAI/GL/wiki/Using_HTML5_nav_element says : 

> "However, in order to make the structure more accessible to user agents that support ARIA as well as ensuring that user agents that don't support HTML5 can also understand the structure, adding the ARIA role="navigation" is advised."

I'm not sure if that is still relevant, but it was about 2 years ago when tested against a range of screen readers.

Will remove it I think, because some automated tests balk at it.

packages/Molecules/Pagination/index.vue
dylankelly: Added this as a prop because you might want to use pagination for items or cards instead of pages.

packages/Molecules/Pagination/index.vue
dylankelly: fixed dat

PR #321, opened 2019-04-08
[SDPA-2341] Move wysiwyg styles into RplMarkup

dylankelly:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2341

### Changed

1.  Move styles from nuxt https://github.com/dpc-sdp/vic-gov-au/tree/develop/modules/nuxt-tide/lib/components/scss/wysiwyg into RplMarkup
1.  Add plugin example for iFrames
1. Add plugin example for embedded media video
1. Add individual stories for
4.1 Link
4.1 Quotation
4.1 Embedded Document
4.1 Tables
4.1 Iframes
4.1 Iframe video embeds
1. Use existing RplEmbeddedVideo component for video embeds


packages/Organisms/Markup/examplePlugins.js
```diff
@@ -1,8 +1,3 @@
-/*
```
alan-cole: Is it still worth having some sort of comment to explain the use of the file?

packages/Organisms/Markup/examplePlugins.js
```diff
@@ -61,37 +64,105 @@ const pluginEmbeddedDocument = function () {
...
+    `
+    return element.replaceWith(RplEmbeddedVideo)
   })
 }
 
 const pluginLinks = function () {
   this.find('a').map((i, el) => {
     const $a = this.find(el)
     const href = $a.attr('href')
+    // const text = htmlUtilities.decodeSpecialCharacters($a.text())
```
alan-cole: I'm guessing this is a nuxt-specific function that we can't use in ripple at the moment. It's good to leave some sort of notification in there that "you'll likely need to decode your html here" - so I can't decide if this comment should stay or go.

PR #324, opened 2019-04-10
[SDPA-1578] Add focus outline to header / footer.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1578

### Changed

1.  Implement white focus ring for dark theme elements (header / footer).
2. Cursor disappears when hovering over footer headers on desktop.
3. Demo data using incorrect url key on links object.

### Screenshots

header
footer


packages/Atoms/Global/scss/tools/_focus.scss
```diff
@@ -0,0 +1,5 @@
+@mixin rpl_focus_dark {
```
dylankelly: 👍

PR #352, opened 2019-05-07
[SDPA-2519] Add specific print url to links.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2519

### Changed

1.  Add specific print url to links.

### Screenshots

Screen Shot 2019-05-07 at 2 41 18 pm
_Screenshot is of a nuxt when these changes are paired with nuxt changes._

Screen Shot 2019-05-07 at 3 01 50 pm
_Ripple and nuxt changes side-by-side. Ripple left, Nuxt right._

### DIFF
--- a/packages/Atoms/Global/index.js
+++ b/packages/Atoms/Global/index.js
@@ -6,6 +6,7 @@ const RplGlobal = {
     const rplOptions = {
       nuxt: false,
       hostname: 'localhost',
+      origin: '', // URL with protocol://host(:port) e.g. http://localhost:3000
       quickexit: false,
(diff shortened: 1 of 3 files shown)


packages/Atoms/Global/index.js
```diff
@@ -6,6 +6,7 @@ const RplGlobal = {
     const rplOptions = {
       nuxt: false,
       hostname: 'localhost',
+      origin: '', // URL with protocol://host(:port) e.g. http://localhost:3000
```
dylankelly: So we'll need to make sure we add both origin and hostname in nuxt then?

PR #354, opened 2019-05-09
[SDPA-2379] Publication download and print block.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2379

### Changed

1.  Publication download and print block.

### Screenshots

Screen Shot 2019-05-09 at 10 31 55-fullpage


packages/Organisms/Publication/PublicationDownloadPrint.vue
```diff
@@ -0,0 +1,114 @@
...
+    DocumentLink
+  },
+  props: {
+    links: Array,
+    showPrint: { type: Boolean, default: true },
+    printText: { type: String, default: 'Print this page' }
+  },
+  methods: {
+    printClick () {
+      window.print()
```
dylankelly: As discussed, can we emit an event on print? Maybe called publicationPrint? That way we can use it for analytics tracking (I know people can invoke the print window separately too).

packages/Organisms/Publication/PublicationDownloadPrint.vue
```diff
@@ -0,0 +1,114 @@
+<template>
+  <ul class="rpl-publication-download-print">
+    <li v-for="(link, index) in links" :key="index">
+      <rpl-document-link class="rpl-publication-download-print__print-item" v-bind="link" />
```
dylankelly: probably need to emit an event here on click too - publicationDownload. With the name of the publication as payload

alan-cole: Added an event.

packages/Organisms/Publication/PublicationDownloadPrint.vue
alan-cole: Added an event.

PR #366, opened 2019-05-16
Added custom icon support for adding new and overriding existing icons.

alan-cole:
@tim-yao - This is my take on a custom icon solution. I'll try implement on nuxt and see if I can update the other branch.


packages/Atoms/Icon/icon-library.js
```diff
@@ -0,0 +1,44 @@
+const iconProps = {}
```
tim-yao: In dev mode, hot reload will run code multiple times if you edited the config file, so this global var value may will be kept instead of reset to `{}`.

alan-cole: Added ability to invoke a reset to these properties and added it before the first custom library is called. This allows require.context() calls be reset in dev hot reloading.

packages/Atoms/Icon/icon-library.js
```diff
@@ -0,0 +1,44 @@
...
+      }
+    }
+  })
+}
+
+function getIconProps () {
+  // Add default icons on first call.
+  // For correct overrides; this must run after custom icons have been added.
+  if (!hasRunOnce) {
+    hasRunOnce = true
```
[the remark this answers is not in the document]
alan-cole: Added ability to invoke a reset to these properties and added it before the first custom library is called. This allows require.context() calls be reset in dev hot reloading.

PR #380, opened 2019-06-03
[SDPA-2629] Remove breakpoint mixin for pagination and set padding

dylankelly:
Make pagination in search result wrap full width and add padding-bottom to account for mobile. 

**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2629

### Changed

1.  Remove conditionally setting pagination steps per breakpoint (breaks in SSR - SDPA-2629)
1.  Reserve space for scroll to top to sit below pagination steps

#### Before

Screen Shot 2019-06-03 at 2 43 22 pm

#### After

Screen Shot 2019-06-03 at 2 42 41 pm

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -934,9 +934,9 @@
       }
     },
     "@types/node": {
-      "version": "12.0.3",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-12.0.3.tgz",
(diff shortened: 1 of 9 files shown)


tim-yao:
@dylankelly as my thought, this makes an extra space added to search result by assume always has no other thing between pagination and back to top. But in real world, the back to top is under "below content" section. So it possible in the future have these issues:
- Search results have that space which not have back to top in it.
- A other component in "below content" section blocked by back to top.

PR #388, opened 2019-06-11
Added ability to pass custom controls as props to map.

kel30a:
The zoom controls are added to the map by default, but I wanted to be able to add other controls such as the FullScreen control (I've added a screenshot of this)

@tim-yao Not sure if I have permissions to add reviewers or labels - but I think this is ready
### Changed
1.  Map.vue

### Screenshots
You can see in this screenshot I've passed in the default OpenLayers FullScreen control and it has rendered the fullscreen button on the map.
image

### DIFF
--- a/packages/Molecules/Map/Map.vue
+++ b/packages/Molecules/Map/Map.vue
@@ -38,11 +38,13 @@ let map,
  */
 const methods = {
   createMap () {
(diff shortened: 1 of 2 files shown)


kel30a:
No longer required.

PR #393, opened 2019-06-18
[SDPSUP-441] Set correct tls-acme setting.

conantran:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPSUP-441

### Changed

1.  Set correct tls-acme setting in .lagoon.yml (`tls-acme: 'false'`)


conantran:
This changes can only be UAT once it's merged in so no QA is needed.

PR #397, opened 2019-06-24
Changed ripple to monorepo for whole Front end

tim-yao:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2505

### Changed

- Added nuxt modules into packages
- Added example
- Added yarn workspaces
- Moved ripple components into packages

Note:
- More work required to make it stable to use.


tim-yao:
CI hasn't been setup so won't work for now.

PR #400, opened 2019-06-25
[SDPA-2776] Lerna update

tim-yao:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2776

### Changed

1.  Setup Lerna with new monorepo changes
2. Published 1.0.0-beta.5
3. Minor fixes for structures

### Screenshots
https://www.npmjs.com/search?q=%40dpc-sdp%2Fripple&page=0&perPage=20


lerna.json
```diff
@@ -1,6 +1,6 @@
 {
-  "lerna": "0.0.0",
-  "version": "1.0.0-beta.1",
+  "version": "1.0.0-beta.5",
+  "exact": true,
```
tim-yao: This make us to use the same exact version across all packages.

package.json
```diff
@@ -1,6 +1,5 @@
 {
   "name": "@dpc-sdp/ripple",
-  "version": "1.0.0-beta.1",
```
tim-yao: Removed it as it won't be updated by Lerna. Babel didn't has that in their monorepo as well.

PR #414, opened 2019-07-05
[SDPA-2833] Update webform conditional.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2833

### Changed

* Re-factored conditional logic code to improve readability.
* Added support for AND / OR / XOR operators.
* Added support for filled, checked, unchecked, value, !value, pattern, !pattern, less, greater triggers.
* Added conditional logic unit tests.

### Screenshots

No screens.

**Note**: This contains a breaking change.

### DIFF
--- a/packages/ripple-nuxt-tide/lib/components/AppForm.vue
+++ b/packages/ripple-nuxt-tide/lib/components/AppForm.vue
@@ -53,7 +53,7 @@ export default {
     checkFieldStates () {
       this.fields.forEach(field => {
         if (field.states) {
-          conditionalLogic(field, field.states, this.formData)
+          conditionalLogic(field, this.formData)
         }
       })
     },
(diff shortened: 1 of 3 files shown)


packages/ripple-nuxt-tide/lib/components/AppForm.vue
```diff
@@ -53,7 +53,7 @@ export default {
     checkFieldStates () {
       this.fields.forEach(field => {
         if (field.states) {
-          conditionalLogic(field, field.states, this.formData)
+          conditionalLogic(field, this.formData)
```
alan-cole: @tim-yao  This is a breaking change, if any other sites are using conditional logic.

tim-yao: @alan-cole only App-form is using it as I can see inside our org projects, so it should be fine.

PR #417, opened 2019-07-05
[SDPA-2730] Publication Image - if no caption, then title always shows.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2730

### Changed

* Publication Image - if no caption, then title always shows.

### Screenshots

Screen Shot 2019-07-05 at 16 13 22-fullpage


packages/ripple-ui-components/components/Organisms/ImageGallery/FullscreenImage.vue
```diff
@@ -40,6 +41,11 @@ export default {
     return {
       toggleCaption: false
     }
+  },
+  computed: {
+    isCaptionShowing () {
+      return (this.caption && this.toggleCaption) || !this.caption
```
tim-yao: I feel this logic is a little confusing... The second part, if there is no caption, then we show it?

alan-cole: In the caption panel exists the title and the caption.

If we have a caption prop
  then we want a button to show / hide the caption panel depending on the toggle state.
If no caption prop was given
  then we want want that caption panel to always display (so we can see the title)

PR #424, opened 2019-07-11
Add back custom component loader

dylankelly:

### Changed

1.  Add back custom component loader, handle config not being present.


packages/ripple-nuxt-tide/lib/core/componentLoader.js
```diff
@@ -1,5 +1,15 @@
...
+let customComponents = {}
+try {
+  const customComponentsLoadConfig = require('~/tide/tide.load-components.js').default
+  if (customComponentsLoadConfig) {
+    customComponents = {
+      ...customComponentsLoadConfig
+    }
+  }
+} catch (error) {
+  // no custom components
```
tim-yao: @dylankelly have you tried this? Will it still throw errors in console? I remembered this catch won't prevent of that.

dylankelly: Yep I tried this without the file and adding the try catch block works fine, I dont think we need to console anything here, we don't need to know if there aren't any custom components

dylankelly:
@tim-yao as discussed there doesn't seem to be a nicer way to implement this conditional loading

PR #425, opened 2019-07-11
[SDPA-2857] Add interface to extend markup plugins.

dylankelly:

### Changed

1.


dylankelly:
No luck with the config loader idea, this should probably just be merged as is

dylankelly:
No luck with the config loader idea, having the try catch is probably safe enough for now

PR #432, opened 2019-07-15
Add unit tests for authenticated content setServerToken

dylankelly:

### Changed

1.  Added unit tests


packages/ripple-nuxt-tide/modules/authenticated-content/__tests__/authenticate.test.js
```diff
@@ -0,0 +1,54 @@
+import { serverSetToken } from './../lib/authenticate'
```
tim-yao: Hey @dylankelly Should we use same name for tests directory across the whole project? At this moment, all are using `test` which is same as vue.js and nuxt.js project. I feel `__tests__` introduce the inconsistency.

packages/ripple-nuxt-tide/modules/authenticated-content/__tests__/authenticate.test.js
```diff
@@ -0,0 +1,54 @@
...
+        }
+      }
+    })
+
+    serverSetToken(`authenticatedContent=${validToken};`, store)
+    expect(store.dispatch).toHaveBeenCalledWith('tideAuthenticatedContent/setAuthenticated', true)
+  })
+
+  it('should isAuth to false when an expired JWT is passed in cookie', () => {
+    MockDate.set('2022-07-15T06:19:17+00:00')
```
tim-yao: 2022 is not very far, the test will fail three years later.

packages/ripple-nuxt-tide/modules/authenticated-content/__tests__/authenticate.test.js
dylankelly: I would like to change this, i think unit tests should always be colocated with corresponding code. I will change this PR and make another which changes this across the project though.

packages/ripple-nuxt-tide/modules/authenticated-content/__tests__/authenticate.test.js
dylankelly: yeah i'll make it one thousand years, by then the earth will be dead anyway 😄

packages/ripple-nuxt-tide/modules/authenticated-content/__tests__/authenticate.test.js
tim-yao: @dylankelly Just check incase you forgot this one...

PR #435, opened 2019-07-16
fix for svgloader adding duplicate config

dylankelly:

### Changed

1.  Fix svgLoader being added twice
1. Resolve project root with `app-root-path` package - https://www.npmjs.com/package/app-root-path - Handles cross platform path resolution


packages/ripple-nuxt-ui/lib/module.js
```diff
@@ -25,16 +25,19 @@ module.exports = function nuxtRipple (moduleOptions) {
...
-      include: [
-        resolve(__dirname, rippleIconPath),
-        resolve(__dirname, `${appDir}/assets/ripple-icon/`)
-      ],
-      use: [
-        'svg-sprite-loader',
-        'svgo-loader'
-      ]
-    })
+    const svgLoader = config.module.rules.find((rule) => rule.use && rule.use.find(r => r === 'svgo-loader'))
```
tim-yao: Should we just remove this duplicate loader from `ripple-nuxt-tide` as it already be added in https://github.com/dpc-sdp/ripple/blob/develop/packages/ripple-nuxt-ui/lib/module.js#L28 ? The `ripple-nuxt-ui` is a dependency of this module.

Issue #77, opened 2018-09-18
What browsers do you aim to support?


rooby:
What browsers do you aim to support?

This would be useful information to add to the contributing docs.

tim-yao:
Hey @rooby Below is the original target made one year ago, we may review them soon.

**Browser Compatibility**
Desktop
Chrome 10+, Firefox 4+, Safari 5+, Opera 12+, IE 10+ 

Mobile
Safari for I0S 3+, Android 2+, Chrome for Android and IOS 4+, Firefox for Android, IE 10+ for Windows 8 devices

As this is a Vue.js project, so the JS supports all browsers that are [ES5-compliant](http://kangax.github.io/compat-table/es5/) (IE8 and below are not supported) 

We will update the doc to add the "Browser Compatibility". Thanks for your help!

PR #442, opened 2019-07-18
[SDPA-2911] Update Readme

dylankelly:

### Changed

1.  Updated Readme formatting.


tim-yao:
@dylankelly any place you think we can add this https://github.com/dpc-sdp/ripple/issues/77 ?

README.md
```diff
@@ -1,26 +1,119 @@
...
+
+- Install [nodejs](https://nodejs.org/en/) (^v10.0.0)
+- Install [yarn](https://yarnpkg.com/en/docs/install) (currently npm is unsupported)
+
+### Install
 
 Run `yarn install`
 
-## Running Example
+### Running Example Application
```
tim-yao: I think this part is belong to 'CONTRIBUTING', as this is for local env setup of core devs work.

Users of Ripple should just want to know how to install Ripple into their project.

README.md
```diff
@@ -1,26 +1,119 @@
...
+
+The example application requires a working [Tide](https://github.com/dpc-sdp/tide) Drupal installation to connect to.
 
 Ensure a populated `.env` exists within the `/examples/vic-gov-au/` directory.
 You can use `/examples/vic-gov-au/example.env` as a template.
 
 Run `yarn run start:example`
 
-## Running Storybook
+### Running Storybook
```
tim-yao: Same as above, I think this is "CONTRIBUTING" doc.

README.md
dylankelly: They will not interact with "Ripple" they will interact with "@dpc-sdp/ripple-nuxt-tide" Which is where that documentation should be. But I have updated the Ripple README.md with basic usage for each package. I think this should be an overview of everything with links for more detail as required.

README.md
dylankelly: See my comment above

PR #466, opened 2019-08-01
[SDPA-2983] Links default to opening in the same window 

dylankelly:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2983

### Changed

1.  Default links to opening in the same window 
1.  Add ripple global option for changing the default behaviour


packages/components/Atoms/Link/Link.vue
```diff
@@ -57,8 +57,8 @@ export default {
     if (!isAnchorLink(this.href) && isRelativeUrl(this.href)) {
       this.isNuxtLink = this.rplOptions.nuxt
     }
-    // Set link target for non nuxt-links
-    if (this.target.length === 0) {
+    // Set link target for non nuxt-links, if global option is true
+    if (this.target.length === 0 && this.rplOptions.externalLinksInNewWindow) {
```
tim-yao: Do we need update doc for this option? https://github.com/dpc-sdp/ripple/blob/develop/packages/components/Atoms/Global/README.md

dylankelly: yep, updated

PR #473, opened 2019-08-05
[SDPA-1407] Make success icon the same size as close on grants overview

dylankelly:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-1407

### Changed

1.  Make success icon the same size as close on grants overview


dylankelly:
@alan-cole @tim-yao  - We need to discuss the icon system, we shouldn't have to mess around like this to size things. I have created a task to come up with a better solution - https://digital-engagement.atlassian.net/browse/SDPA-3006

PR #478, opened 2019-08-07
[SDPA-2594] Hero Banner background position.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2594

### Changed

* Use grid and layout variables to calculate background position.
* Move scss variables into an array / map to reduce repetition.

The background position uses these rules as guidelines:
* The graphic should at least reach the right edge of the window.
* The graphic should sit 100px underneath the orange box (unless doing so breaks rule 1).
* The top of the graphic edge should touch the bottom right of the orange box (Assuming the orange box sits 68px from the bottom of the hero banner).

### DIFF
--- a/packages/components/Organisms/HeroBanner/HeroBanner.vue
+++ b/packages/components/Organisms/HeroBanner/HeroBanner.vue
@@ -165,34 +165,60 @@ export default {
     @include rpl_mobile_padding;
     position: relative;
     background-repeat: no-repeat;
-    background-position: right -3.75rem bottom -1rem;
+
+    // ---------
+    $hero-banner-grid-container-gutter-s: 20px;
+    $hero-banner-grid-container-gutter-m: 32px;
+    $hero-banner-grid-container-gutter-l: 68px;
+    $hero-banner-grid-container-gutter-xl: '((100vw - 1056px)/2)';
+    $hero-banner-links-padding-xl: rpl_grid_get_gutter($rpl-grid) / 2;
+    $hero-banner-bg-links-overlap: 100px;
+    $hero-banner-bg-vert-offset-xs: ($rpl-space * -5);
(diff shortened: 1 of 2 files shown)


packages/components/Organisms/HeroBanner/HeroBanner.vue
```diff
@@ -160,39 +160,66 @@ export default {
...
+    's': ('pos': 'edge', 'size': 20rem, 'gutter': $rpl-component-padding-xs, 'v-offset': rem(-16px)),
+    'm': ('pos': 'edge', 'gutter': map-get(map-get($rpl-layout, 'site_padding'), 'm')),
+    'l': ('pos': 'edge', 'gutter': map-get(map-get($rpl-layout, 'site_padding'), 'l')),
+    'xl': ('pos': 'edge', 'gutter': $rpl-hero-banner-background-graphic-gutter-xl),
+    '1490px': ('pos': 'right'),
+    'xxl': ('pos': 'edge', 'size': 40rem, 'gutter': $rpl-hero-banner-background-graphic-gutter-xl, 'h-offset': rem(33px), 'v-offset': $rpl-hero-banner-background-graphic-v-offset-xxl),
+    '2072px': ('pos': 'right', 'v-offset': $rpl-hero-banner-background-graphic-v-offset-xxl),
+    'xxxl': ('pos': 'right', 'v-offset': $rpl-hero-banner-background-graphic-v-offset-xxl),
+    '2660px': ('pos': 'right')
+  ) !default;
```
tim-yao: This is really a complex code solution for a design issue. I would suggest review the design instead of adding complex logic into component as it will be hard to add new variations in the future. I have checked the result http://storybook-ripple-pr-478.lagoon.vicsdp.amazee.io/?path=/story/organisms-herobanner--default which is working well. I will pass it, cheers!

PR #479, opened 2019-08-08
[SDPA-910] Add padding to Display Heading typography.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-910

### Changed

* Add padding to Display Heading typography.
* Update typography function to optionally include base rules (rules that are only needed once for all breakpoints).

### Screenshots

Screen Shot 2019-08-08 at 2 51 11 pm

### DIFF
--- a/packages/components/Atoms/Global/scss/tools/_typography.scss
+++ b/packages/components/Atoms/Global/scss/tools/_typography.scss
@@ -1,4 +1,4 @@
-@mixin rpl_typography_font($font-size: 's', $line-height: 1.2em, $font-family: 'inherit') {
+@mixin rpl_typography_font($font-size: 's', $line-height: 1.2em, $font-family: 'inherit', $include-base: true) {
   @if $font-family != 'inherit' {
     font-family: rpl-font($font-family);
   }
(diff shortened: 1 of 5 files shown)


packages/components/Atoms/Global/scss/tools/_typography.scss
```diff
@@ -9,17 +9,32 @@
...
 }
 
-@mixin rpl_display_background($font-size) {
-  background-color: rpl-color('secondary');
-  color: rpl-color('white');
-  display: inline;
+@mixin rpl_display_background($font-size, $include-base: true) {
+  @if $include-base == true {
+    background-color: rpl-color('secondary');
+    color: rpl-color('white');
```
tim-yao: Let's use this PR to move hardcoded color out of this mixin, it's hard to customised. I will suggest to use sass variables like `$rpl-display-background-color` and `$rpl-display-text-color`.
We should never use `rpl-color` directly, always use variable.

packages/components/Atoms/Global/scss/tools/_typography.scss
```diff
@@ -1,4 +1,4 @@
-@mixin rpl_typography_font($font-size: 's', $line-height: 1.2em, $font-family: 'inherit') {
+@mixin rpl_typography_font($font-size: 's', $line-height: 1.2em, $font-family: 'inherit', $include-base: true) {
```
tim-yao: Hi @alan-cole Any purpose for `include-base` parameter? Looks like no place is not using it?

tim-yao:
Also, can we add a notes in somewhere docs, stories, for no IE11/Edge support on padding?

packages/components/Atoms/Global/scss/tools/_typography.scss
alan-cole: Will fix.

packages/components/Atoms/Global/scss/tools/_typography.scss
alan-cole: This is for line 63 - 68 of this file:

> Avoid unnecessary duplication; include base styles only on first breakpoint.

When we use the breakpoint loop in rpl_typography_ruleset() for adding styles; only the font-family / line-height / font-size change, but we are also repeating all the other rules as well. This cuts down on some of that unnecessary duplication.

tim-yao: @alan-cole that makes sense, thanks!

PR #486, opened 2019-08-09
[SDPA-372] Add Topic and Tags to end of Landing Page.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-372

### Changed

1.  Add Topic and Tags to end of Landing Page.

NOTE: This has been implemented before the back-end changes needed for the topic URL, therefore it’s built on the assumption that the slug url will be found in path.alias. This needs to be tested after that BE change is implemented.

### Screenshots

Screen Shot 2019-08-09 at 17 02 30-fullpage


packages/ripple-nuxt-tide/modules/landing-page/pages/index.vue
```diff
@@ -38,6 +43,24 @@ export default {
...
+        let metaTags = []
+        if (this.page.field_topic) {
+          metaTags.push(this.page.field_topic)
+        }
+        if (this.page.field_tags) {
+          metaTags = metaTags.concat(this.page.field_tags)
+        }
+        return metaTags.map(tag => ({
+          text: tag.name,
+          url: (tag.path && tag.path.alias) ? tag.path.alias : ''
```
tim-yao: Not sure what tag url Tide will pass to us. If it's a Nuxt route path instead of CMS path, I think we will need some prefix e.g `/tag/` so we can handle it in Nuxt. @alan-cole @dylankelly

alan-cole: I had discussed with Sonny before implementing. He said it would be the full path including the prefix.

alan-cole: I have tested this with https://nginx-php-content-vic-pr-649.lagoon.vicsdp.amazee.io (SDPA-2987). It works as expected.

PR #487, opened 2019-08-12
[SDPA-2976] Fetch alert notifications from site taxonomy

dylankelly:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2976

### Changed

1.  Removes seperate call for alerts and gets them from site taxonomy now.
1.  Multiple alerts can be present now.


dylankelly:
@tim-yao I've added a router middleware to check when the sitedata was last fetched and refetch it if its been a while (set to 1 minute but we can tweak it). I reckon thats a good compromise. What do you think?

packages/ripple-nuxt-tide/modules/alert/alert-middleware.js
```diff
@@ -0,0 +1,17 @@
+import { get } from 'lodash'
```
tim-yao: So this is actually refresh site data, not only alert. I would like to move this to site module, how do you think?

packages/ripple-nuxt-tide/lib/core/tide.js
```diff
@@ -108,57 +108,60 @@ export const tide = (axios, site, config) => ({
...
-    if (sitesData instanceof Error) {
-      return sitesData
+    if (siteId !== null) {
+      params.filter = {
+        drupal_internal__tid: {
+          path: 'drupal_internal__tid',
+          value: siteId
+        }
+      }
+      const response = await this.get(`taxonomy_term/sites`, params)
```
tim-yao: This will get `undefined` if API request failed. I need to change this to handle error on each request instead of handle it globally in https://github.com/dpc-sdp/ripple/blob/develop/packages/ripple-nuxt-tide/lib/templates/plugin.js#L61. As the error been resolved there to prevent break the site.

What we need to do now, is check if it's `undefined` before pass into `jsonapiParse.parse`.

tim-yao: Also, any reason we didn't use `getSitesData()`? We need to handle the pagination as well.

dylankelly: as discussed, getSiteData was fetching ALL sites and then just picking the needed one - have changed this to only fetching the needed one using filter param

examples/vic-gov-au/package.json
```diff
@@ -11,7 +11,7 @@
     "cy:open": "cypress open",
     "cy:run": "cypress run -e TAGS='not @skip or @smoke'",
     "cy:run-smoke": "cypress run -e TAGS='@smoke' --spec 'test/e2e/integration/smoke/*'",
-    "test:dev": "cross-env BASIC_AUTH=0 start-server-and-test start:build http://localhost:3000 cy:open",
+    "test:dev": "start-server-and-test dev http://localhost:3000 cy:open",
```
tim-yao: Any reason use dev to test, instead of build version?

packages/ripple-nuxt-tide/lib/core/middleware.js
```diff
@@ -189,12 +189,18 @@ export default async function (context, results) {
           }
         }
         // site section nav
-        if (results.tidePage.field_show_site_section_nav) {
-          const addSectionNavMenu = await context.app.$tide.getSiteData(results.tidePage.section).then(appSectionData => {
+        if (results.tidePage.section) {
+          const addSectionNavMenu = await context.app.$tide.getSiteData(results.tidePage.section).then(async siteData => {
+            // save alerts if site section has them
+            if (context.app.$tide.isModuleEnabled('alert')) {
```
tim-yao: I would suggest update here a little but it's ok to pass based on the time pressure.
The original code(from Guy) is to `addSectionNavMenu`, so insert `alert` related logic into it is not right for me. We can get the sitedata first then do the two logic work after that.

packages/ripple-nuxt-tide/lib/pages/Tide.vue
```diff
@@ -128,6 +128,9 @@ export default {
     if (this.page.sidebarComponents && this.page.sidebarComponents.length > 0) {
       this.sidebarComponents = this.$tide.getDynamicComponents(this.page.sidebarComponents, this.page.sidebar)
     }
+    if (this.page.section) {
+      this.$store.commit('tideSite/setSiteSection', parseInt(this.page.section, 10))
```
tim-yao: I feel this is not same as the way you talked yesterday... this will only work in Tide page, what about other static route pages, e.g search?

examples/vic-gov-au/package.json
dylankelly: In production build you cant load dev tools for debugging - this command is just for running test locally and interactively

packages/ripple-nuxt-tide/lib/pages/Tide.vue
dylankelly: I agree - but I think we should add a ticket for refactoring middleware into composable functions and handle it there

PR #503, opened 2019-08-23
[SDPA-2965] Accordion print styling.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2965

### Changed

* Add slight padding to accordions.
* Update accordion text / icon colours for print.
* Accordion icon to always appear downwards on print.
* Fix MetaTag border colour for print.

### Screenshots

Screen Shot 2019-08-23 at 12 27 48 pm

### DIFF
--- a/packages/components/Organisms/Accordion/Accordion.vue
+++ b/packages/components/Organisms/Accordion/Accordion.vue
@@ -163,6 +163,10 @@ export default {
       margin: 0;
       padding: 0;
       background-color: $rpl-accordion-background-color;
+
+      @media print {
+        margin: $rpl-space-4 0;
+      }
     }
 
     &__list-item {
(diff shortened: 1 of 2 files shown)


packages/components/Organisms/Accordion/Accordion.vue
```diff
@@ -196,7 +204,7 @@ export default {
 
     &__button {
       @include rpl_typography_ruleset($rpl-accordion-title-ruleset);
-      color: $rpl-accordion-button-text-color;
+      @include rpl_text_color($rpl-accordion-button-text-color);
```
tim-yao: Hi @alan-cole should we always use `rpl_text_color ` instead of `color` in Ripple?

alan-cole: Basically, yes. Any text that is part of the printout should use rpl_text_color, as it includes additional @media print color definition.

PR #508, opened 2019-08-28
[SDPA-3128] Added property to exclude hits.

GROwen:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-3128

### Changed

1.  Added a new property to the search options to exclude combinations of content-types and records with fields values from search results

### DIFF
--- a/packages/ripple-nuxt-tide/modules/search/pages/TideSearch.vue
+++ b/packages/ripple-nuxt-tide/modules/search/pages/TideSearch.vue
@@ -83,7 +83,11 @@ export default {
           'title',
           'type',
           'url'
-        ]
+        ],
+        exclude: {
+          'type': 'grant',
+          'field': 'field_node_link'
+        }
       }
     }
   },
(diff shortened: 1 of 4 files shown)


packages/ripple-nuxt-tide/modules/search/pages/TideSearch.vue
```diff
@@ -83,7 +83,11 @@ export default {
           'title',
           'type',
           'url'
-        ]
+        ],
+        exclude: {
```
GROwen: I feel like this should be conditionally set based on the status of the nuxt-tide Grant sub-module. Is that required?

tim-yao: Hi @GROwen I think this is depending on how we design the search index in backend and hard to tell just in front end. I feel the best way to control in front end is extensible whitelist. The `exclude` is a blacklist, but still be ok here I think. We don't have the ability to extend the config for this default search page for now, so we can't move this setting into Grant module. Unless we add a API/hook to allow custom config merging into this search config.

But you need to test and make sure, without `Grant` enabled in CMS, this exclude query is still working.

GROwen: @tim-yao Yep the query will still return results with an absence of any docs of `type: grant` in an index.

A whitelist won't be possible because of the additional requirement to check for non-null values - which complicates things. I also didn't want to go overboard with the flexibility of this solution because we don't know of any other requirements and it may all be made redundant by the UX discovery and refactor of the search implementation.

PR #528, opened 2019-09-11
[SDPA-2949] Fix: Search results query to support filter object.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2949

### Changed

* setFiltersOnCreate() did not support the current URL structure.


tim-yao:
@alan-cole Are we able to add a unit test for what this fix has solved?

PR #529, opened 2019-09-12
[SDPA-3117] Fixes to anchor link.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-3117

### Changed

* Remove duplicate link functionality - introduces too much complexity.
* Include a single function getAnchorLinkName to produce all names.
* Use this function in all situations.
* Update tests for function.
* Check for valid selector on load before scrolling.
* Anchors will correctly scroll if navigated to from a page link.


src/test/unit/helpers.test.js
```diff
@@ -50,3 +50,16 @@ describe('isAnchorLink', () => {
...
   /* eslint-enable indent */
 })
+
+describe('getAnchorLinkName', () => {
+  /* eslint-disable indent */
+  test.each`
+    text                                                | expected
+    ${'Anchor&nbsp;&amp;&nbsp;&copy;&nbsp;Link'}        | ${'anchor-link'}
+    ${'Anchor`~!@#$%^&*()-_=+{}[]\\|;:\'"<>,.?/\nLink'} | ${'anchor-link'}
+    ${'   ANCHOR   LINK   '}                            | ${'anchor-link'}
```
tim-yao: I would suggest add one more test case on title include numbers.

packages/ripple-nuxt-tide/lib/module.js
```diff
@@ -33,6 +33,33 @@ const nuxtTide = function (moduleOptions) {
     fileName: './request-id.js'
   })
   this.options.router.middleware.push('request-id')
+
+  // https://toor.co/blog/nuxtjs-smooth-scrolling-with-hash-links/
+  this.options.router.scrollBehavior = async (to, from, savedPosition) => {
```
tim-yao: Would like to know what problem this fixed, I have a fix for the scroll already, is that issue still there?

PR #539, opened 2019-09-23
[SDPA-3216] CTA description fix for special characters

MdNadimHossain:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-3216

### Changed

1.  Added new filter
2. Updated CTA summary field to use the new filter
3. Updated html-utilities.js to capture more special characters


packages/ripple-nuxt-tide/lib/core/html-utilities.js
```diff
@@ -5,7 +5,9 @@ export default {
       '&gt;': '>',
       '&lt;': '<',
       '&apos;': "'",
-      '&nbsp;': ' '
+      '&nbsp;': ' ',
+      '&quot;': '"',
+      '&#039;': "'"
```
tim-yao: Hi @MdNadimHossain can you please update this test for `&#039;`? https://github.com/dpc-sdp/ripple/blob/develop/packages/ripple-nuxt-tide/test/unit/html-utilities.test.js

MdNadimHossain: Done, added test to check these three - &apos;&quot;&#039;

PR #545, opened 2019-09-25
[SDPA-3184] Setup timeout for Tide request

tim-yao:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-3184

### Changed

1.  Set timeout of Tide API request to 4000ms
2. Moved `axios` logic from default plugin into a separated file for better structure
3. Moved Tide API response error handling to each request instead of handled globally. To make it more flexible, now each function use Tide Get request can do different treatment on error based on their own requirement.
4. Improved error logging and handling so site fail will provide better informations for debugging.


packages/ripple-nuxt-tide/test/unit/tide.test.js
```diff
@@ -145,11 +145,13 @@ describe('tide', () => {
...
     mockAxios.$get.mockRejectedValue(error)
 
-    expect.assertions(1)
     const pathNotExist = '/no-such-page'
-    await tideApi.getPathData(pathNotExist).catch(e => {
-      expect(e.response.status).toBe(404)
-    })
+    try {
+      await tideApi.getPathData(pathNotExist)
+      expect(true).toBe(false)
```
alan-cole: Perhaps this could have a single test to see if catch statement ran or didn't.

PR #554, opened 2019-10-02
Added different timeout for CI test

tim-yao:

### Changed

1.  Added a request in CI to awake dev BE before e2e test
2. Increased timeout for CI tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,8 +37,14 @@ jobs:
       - persist_to_workspace:
           root: ~/app
           paths: .
+
+  warm-up:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -37,6 +37,10 @@ jobs:
       - persist_to_workspace:
           root: ~/app
           paths: .
+      - run:
```
dylankelly: Might be better to make this a seperate job so it can run first thing (we can choose which workflows it runs in too, we probably only need it to run for nightly workflow)

PR #560, opened 2019-10-08
[SDPA-3209] Added error capture and UI handling

tim-yao:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-3209

### Changed

1.  Added mixins for error capture
2. Added Ripple dev mode option
2. Added a component to display captured error in dev mode only
3. Added error capture on RplCol and several components
4. Let card to throw an error when invalid link provided
5. Fixed the issue mentioned in the ticket

### Screenshots

Screen Shot 2019-10-08 at 3 59 13 pm
Screen Shot 2019-10-08 at 3 00 24 pm


packages/components/Atoms/Grid/__tests__/col.test.js
```diff
@@ -0,0 +1,51 @@
...
+})
+
+const colCatchErrorNotDev = mount(RplCol, {
+  propsData: {
+    catchChildError: true
+  }
+})
+
+// Give an error for testing.
+// Ideally will be mount an error child component but Vue test utils didn't support mount in slot.
```
dylankelly: https://vue-test-utils.vuejs.org/api/options.html#slots

packages/components/Atoms/Global/mixins/catch-child-error.js
```diff
@@ -0,0 +1,70 @@
...
+ * If there is no error captured, no class will be added, otherwise it will add "rpl-child-component-error".
+ * You can then apply your own style when error captured.
+ *
+ *   <div class="my-component" :class="childErrorClass">
+ */
+
+const catchChildError = {
+  data: function () {
+    return {
+      gotChildError: false,
```
dylankelly: hasChildError

packages/components/Atoms/Grid/__tests__/col.test.js
tim-yao: `slots` is ok but it won't mount the child. Here is the issue comment from the vue test utils author:
> There isn't currently a way to pass mounted components to slots...
> We aren't going to implement this at the moment. You can pass a component, but not an instance.
https://github.com/vuejs/vue-test-utils/issues/41#issuecomment-327141769 
I did some try from others solution, but I couldn't get error thrown from child in tests.

tim-yao:
> Approved - but it seems hacky to make all child card components have to implement something. Shouldn't you being doing the error checking in the parent and then just render an error component if required?

@dylankelly No need child component do anything, and parent is optional to use or not use the error capture mixins depends on if they need to capture.

What I did to cards, is let them throw an error so it can be captured by parent when card required link is not provided. Previously card will just not be rendered by no error, in that case we have no way to handle them.

packages/components/Atoms/Global/mixins/catch-child-error.js
tim-yao: The reason use `got` instead of `has` here is it can has child error but if we didn't enable the catch error option, it won't get the error. The child error will then continually go up to next parent.

PR #582, opened 2019-10-27
[SDPA-3310] Whatson event listing filtering by end date

MdNadimHossain:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-3310

### Changed

1.  Added default filter value for end date.

### Screenshots
Screen Shot 2019-10-28 at 9 41 55 am

### DIFF
--- a/packages/ripple-nuxt-tide/modules/event/pages/search.vue
+++ b/packages/ripple-nuxt-tide/modules/event/pages/search.vue
@@ -107,8 +107,7 @@ export default {
           type: 'date',
           values: setFilterDate.endOf('day').toISOString()
         }
-      }
-      else {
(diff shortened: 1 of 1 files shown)


packages/ripple-nuxt-tide/modules/event/pages/search.vue
```diff
@@ -107,6 +107,14 @@ export default {
           values: setFilterDate.endOf('day').toISOString()
         }
       }
+      else {
+        const today = moment().endOf('day').toISOString()
```
tim-yao: Please add timezone on the date, as a user browser may not in the same timezone as Victoria.

PR #597, opened 2019-11-13
CI fix

anthony-malkoun:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-3272

### Changed

1.  Using 8.7.9 version of Drupal


.circleci/config.yml
```diff
@@ -227,21 +228,21 @@ workflows:
           filters:
             branches:
               only:
-                - release/*
+                - /^release\/.*/
```
anthony-malkoun: @dylankelly just and FYI, it's not actually running the full test suites against release branches as the regex was not quite right.

.circleci/config.yml
```diff
@@ -92,12 +92,13 @@ jobs:
...
 
   e2e-vic-gov:
-    # This runs the full suite of e2e tests 
+    # This runs the full suite of e2e tests
     docker:
       - image: cypress/base:10
         environment:
           TERM: xterm
           TZ: "Australia/Melbourne"
+          CONTENT_API_SERVER: "https://nginx-php-content-vic-pr-738.lagoon.vicsdp.amazee.io/"
```
anthony-malkoun: DO NOT MERGE. This line is only to point CI at specific backend

anthony-malkoun:
@tim-yao while this is not required for Drupal 8.7.9 testing there is still a fix for circle config that should go in.

PR #603, opened 2019-11-21
[SDPL-169] Adds border radius to option group

dylankelly:

### Changed

1. Add border-radius to first and last item in option group at Desktop BP
1.  Fix clear button text color
1. Fix export for list-group item

### DIFF
--- a/packages/components/Molecules/Form/fields/fieldRploptionbutton.vue
+++ b/packages/components/Molecules/Form/fields/fieldRploptionbutton.vue
@@ -42,6 +42,7 @@ export default {
   $rpl-option-button-label-color: rpl-color('extra_dark_neutral') !default;
   $rpl-option-button-label-border-width: 1px !default;
   $rpl-option-button-label-border: $rpl-option-button-label-border-width solid rpl-color('mid_neutral_1') !default;
+  $rpl-option-button-label-border-radius: 4px;
   $rpl-option-button-label-checked-border: $rpl-option-button-label-border-width solid rpl-color('primary') !default;
   $rpl-option-button-text-focused-color: rpl-color('primary') !default;
   $rpl-option-button-text-focused-border: $rpl-option-button-label-border-width solid rpl-color('primary') !default;
@@ -80,6 +81,15 @@ export default {
 
         #{$root}__text {
           text-decoration: none;
+        }
+      }
+
+      @include rpl-breakpoint('l') {
+        &:first-child {
+          border-radius: $rpl-option-button-label-border-radius 0 0 $rpl-option-button-label-border-radius;
+        }
+        &:last-child {
+          border-radius: 0 $rpl-option-button-label-border-radius $rpl-option-button-label-border-radius 0;
         }
       }
     }
--- a/packages/components/Molecules/Form/fields/fieldRploptionbutton.vue
+++ b/packages/components/Molecules/Form/fields/fieldRploptionbutton.vue
@@ -15,7 +15,7 @@ export default {
   mixins: [abstractField],
   data () {
     return {
-      optionValues: this.schema.optionValues || [ 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z' ]
+      optionValues: this.schema.optionValues || 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
     }
   },
   computed: {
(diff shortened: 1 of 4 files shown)


packages/components/Molecules/Form/fields/fieldRploptionbutton.vue
```diff
@@ -42,6 +42,7 @@ export default {
   $rpl-option-button-label-color: rpl-color('extra_dark_neutral') !default;
   $rpl-option-button-label-border-width: 1px !default;
   $rpl-option-button-label-border: $rpl-option-button-label-border-width solid rpl-color('mid_neutral_1') !default;
+  $rpl-option-button-label-border-radius: 4px;
```
tim-yao: Hey @dylankelly I think you can reuse global setting `$rpl-form-element-border-radius` so all form element border radius are in same style. https://github.com/dpc-sdp/ripple/blob/3358427e782379eae15f500b0a8aaa34a30169f1/packages/components/Molecules/Form/scss/_form.scss#L10

packages/components/Molecules/Form/fields/fieldRploptionbutton.vue
```diff
@@ -42,6 +42,7 @@ export default {
...
   $rpl-option-button-label-checked-border: $rpl-option-button-label-border-width solid rpl-color('primary') !default;
```
tim-yao: Global form setting `$rpl-form-element-border-color` also can be used https://github.com/dpc-sdp/ripple/blob/3358427e782379eae15f500b0a8aaa34a30169f1/packages/components/Molecules/Form/scss/_form.scss#L8

packages/components/Molecules/Form/fields/fieldRploptionbutton.vue
```diff
@@ -42,6 +42,7 @@ export default {
   $rpl-option-button-label-color: rpl-color('extra_dark_neutral') !default;
```
tim-yao: Hi @dylankelly I think this `$rpl-form-element-text-color` can be used to set your form element text color https://github.com/dpc-sdp/ripple/blob/3358427e782379eae15f500b0a8aaa34a30169f1/packages/components/Molecules/Form/scss/_form.scss#L11

packages/components/Molecules/Form/fields/fieldRploptionbutton.vue
dylankelly: ✅

packages/components/Molecules/Form/fields/fieldRploptionbutton.vue
dylankelly: ✅

packages/components/Molecules/Form/fields/fieldRploptionbutton.vue
dylankelly: ✅

dylankelly:
Hey @tim-yao, have fixed those vars. I've also fixed up the label reference, the labels weren't joined up with the inputs correctly. But it looks like vue-form-generator isnt  doing it right with its own core default radio group component https://github.com/vue-generators/vue-form-generator/blob/master/src/fields/core/fieldRadios.vue It needs to have an ID that matches the label for attribute. If I get around to it i'll submit a PR to VFG. FYI VFG hasn't had any activity for a loooong time.

Current -
Screen Shot 2019-11-22 at 10 44 46 am


Should be

Screen Shot 2019-11-22 at 10 44 28 am


Notice the id on field-wrap and the numbered label input pairs

tim-yao:
> Hey @tim-yao, have fixed those vars. I've also fixed up the label reference, the labels weren't joined up with the inputs correctly. But it looks like vue-form-generator isnt doing it right with its own core default radio group component https://github.com/vue-generators/vue-form-generator/blob/master/src/fields/core/fieldRadios.vue It needs to have an ID that matches the label for attribute. If I get around to it i'll submit a PR to VFG. FYI VFG hasn't had any activity for a loooong time.
> 
> Screen Shot 2019-11-22 at 10 47 51 am
> 
> Current -
> Screen Shot 2019-11-22 at 10 44 46 am
> 
> Should be
> 
> Screen Shot 2019-11-22 at 10 44 28 am
> 
> Notice the id on field-wrap and the numbered label input pairs

Thanks @dylankelly . We can try submit a PR to VFG and see if they will take it. If they stop to maintain it, that's bad news.
For radios `name`, we need to keep same within one group, otherwise BE(Drupal webform) won't able to recognise the dynamic form data name. Same as the new OptionButton I think, even it's not post back to Drupal webform, it may will be an issue in handle the submission.
```<fieldset>
<legend>Output format</legend>
  <div>
    <input type="radio" name="format" id="txt" value="txt" checked>
    <label for="txt">Text file</label>
  </div>
  <div>
    <input type="radio" name="format" id="csv" value="csv">
    <label for="csv">CSV file</label>
  </div>
  […]
</fieldset>```

PR #605, opened 2019-11-28
[SDPA-3491] publication author field bug fix.

MdNadimHossain:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-3491

### Changed

1.  Added the type array for author prop.
2. changed to get the correct array value of author names from the field.
3. Did the same change for the print page.
4. Changed the prop type in the storybook to reflect it. 

### Screenshots
Publication Page - 
Screen Shot 2019-11-28 at 12 40 37 pm
Print all page - 
Screen Shot 2019-11-28 at 12 40 55 pm

### DIFF
--- a/packages/ripple-nuxt-tide/modules/publication/tide.config.js
+++ b/packages/ripple-nuxt-tide/modules/publication/tide.config.js
@@ -8,10 +8,10 @@ module.exports = {
       'field_landing_page_contact.field_paragraph_social_media',
       'field_license_type',
       'field_landing_page_component',
-      'field_publication_authors',
       'field_landing_page_component.field_paragraph_accordion',
       'field_landing_page_component.field_complex_image_media',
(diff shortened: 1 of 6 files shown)


packages/ripple-nuxt-tide/modules/publication/tide.config.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
       'field_landing_page_contact.field_paragraph_phones',
       'field_landing_page_contact.field_paragraph_social_media',
       'field_license_type',
-      'field_landing_page_component',
```
tim-yao: So this can be removed?

MdNadimHossain: nope, did that by mistake, I have added it back.

PR #625, opened 2020-01-15
[SDPA-3649] Added headers to help ops

tim-yao:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-3649

### Changed

1.

### DIFF
--- a/packages/ripple-nuxt-tide/lib/core/tide.js
+++ b/packages/ripple-nuxt-tide/lib/core/tide.js
@@ -51,7 +51,8 @@ export const tide = (axios, site, config) => ({
       timeout: axiosTimeout,
       headers: {
         'Content-Type': 'application/vnd.api+json; charset=UTF-8',
-        'Accept': 'application/vnd.api+json'
+        'Accept': 'application/vnd.api+json',
+        'X-SDP-REQUEST-LOCATION': 'tide'
       }
     }
 
--- a/packages/ripple-nuxt-tide/lib/core/tide.js
+++ b/packages/ripple-nuxt-tide/lib/core/tide.js
@@ -52,12 +52,13 @@ export const tide = (axios, site, config) => ({
       timeout: axiosTimeout,
       headers: {
         'Content-Type': 'application/vnd.api+json; charset=UTF-8',
-        'Accept': 'application/vnd.api+json'
+        'Accept': 'application/vnd.api+json',
+        [RPL_HEADER.REQ_LOCATION]: 'tide'
       }
     }
 
-    // Set request location to tide
-    axiosConfig.headers[RPL_HEADER.REQ_LOCATION] = 'tide'
+    // // Set request location to tide
+    // axiosConfig.headers[RPL_HEADER.REQ_LOCATION] = 'tide'
 
     if (this.isModuleEnabled('authenticatedContent')) {
       // Set 'X-Authorization' header if authToken present
@@ -80,7 +81,7 @@ export const tide = (axios, site, config) => ({
       timeout: 9000,
       headers: {
         'Content-Type': 'application/vnd.api+json;charset=UTF-8',
-        'X-Sdp-Request-Location': 'tide',
+        [RPL_HEADER.REQ_LOCATION]: 'tide',
         'X-Request-Id': helper.generateId()
       }
     }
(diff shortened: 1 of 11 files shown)


packages/ripple-nuxt-tide/lib/module.js
```diff
@@ -11,8 +11,22 @@ const nuxtTide = function (moduleOptions) {
     ...this.options.proxy,
     // Set the proxy timeout for requesting to Tide API as 9 seconds.
     // POST request to Tide normally need more than 5 seconds to get response.
-    '/api/v1/': { target: options.baseUrl, proxyTimeout: 9000 },
-    '/sites/default/files/': options.baseUrl
+    '/api/v1/': {
+      target: options.baseUrl,
+      proxyTimeout: 9000,
+      onProxyRes (proxyRes, req, res) {
+        proxyRes.headers['X-SDP-APP-TYPE'] = 'tide'
```
kurtfoster: @tim-yao , I realised an issue with this. You might need an if before it like this to set the var

        let cacheHeader = 'tide';
        if (response.headers['x-sdp-app-type'] !== undefined) {
          cacheHeader = response.headers['x-sdp-app-type'];
        }
....
        proxyRes.headers['X-SDP-APP-TYPE'] = cacheHeader

Or you might need to move this code down to packages/ripple-nuxt-tide/lib/templates/axios.js, I'm not sure. But this should pull and pass through the X-SDP-APP-TYPE header from the API it's proxying rather than hard code the value, defaulting to 'tide'.

packages/ripple-nuxt-tide/lib/templates/axios.js
```diff
@@ -12,6 +12,30 @@ export default function ({ $axios, app, res }) {
     }
   })
 
+  $axios.onResponse(response => {
+    // Pass Drupal content page response section cache tag to Nuxt page response.
+    // Below content endpoints are hard coded as not need to be extensible for now.
+    const contentReqPatterns = [
+      '/api/v.*/node.*',
```
kurtfoster: If it's not too hard can this pull from nuxt.config.json on the project level instead? Also, I think the menu api all might need to get added here?

packages/ripple-nuxt-tide/lib/templates/axios.js
```diff
@@ -12,6 +12,30 @@ export default function ({ $axios, app, res }) {
...
+    const reqRegexStr = '^(?:' +
+      contentReqPatterns.join('|') +
+      ')$'
+    const contentReqRegex = new RegExp(reqRegexStr)
+    const reqPath = response.request.path
+
+    if (reqPath !== undefined && res) {
+      if (reqPath.match(contentReqRegex)) {
+        if (response.headers['section-cache-tags'] !== undefined) {
+          res.setHeader('Section-Cache-Tags', response.headers['section-cache-tags'])
```
kurtfoster: This is a bit anal, but can we move all the headers to lower case to keep them consistent.

packages/ripple-nuxt-tide/lib/module.js
steveworley: Will Drupal ever proxy an API directly? My understanding is that `/api/v1` is set up to proxy Drupal directly. Requests that go to elastic don't go via Drupal they go direct to elastic. So could this ever be anything other than tide?

tim-yao: @steveworley that's what I thought, and why I just hard code `tide` here. As my understanding, `/api/v1/` is pointed to our Drupal Tide API, so should always be `tide`.

Any other proxy responses, if they need go through our Nuxt proxy, then we will set the value in each of proxy config, like I did for `/sites/default/files/` below: https://github.com/dpc-sdp/ripple/pull/625/files/843c3f02ed6c85db9db6238fe514fd8681d678ac#diff-2538535bb7172f81b6518c4233db3ca2R27

Say we need proxy `/mailgun-api/` then we will have a separated proxy config for it. And set 'X-SDP-APP-TYPE' to `mailgun`.

@kurtfoster that's what I was talking about before. Any thoughts?

packages/ripple-nuxt-tide/lib/templates/axios.js
tim-yao: will see if I can add it as a custom config.

packages/ripple-nuxt-tide/lib/core/tide.js
```diff
@@ -76,6 +80,7 @@ export const tide = (axios, site, config) => ({
       timeout: 9000,
       headers: {
         'Content-Type': 'application/vnd.api+json;charset=UTF-8',
+        'X-Sdp-Request-Location': 'tide',
```
steveworley: If we're building with ES6 and Babel can we do?

```
headers: {
  [RPL_HEADER.REQ_LOCATION]: 'tide'
}
```

tim-yao: Hey @steveworley yes, I missed it, will fix it.

PR #637, opened 2020-02-25
Search checkbox filter

tim-yao:

### Changed

1.  to make checkbox search filter working

### DIFF
--- a/packages/ripple-nuxt-tide/modules/search/lib/search/services/elasticsearch/lib/querybuilder.js
+++ b/packages/ripple-nuxt-tide/modules/search/lib/search/services/elasticsearch/lib/querybuilder.js
@@ -39,7 +39,7 @@ const addFilter = function (esbResult, filter, filterName) {
       if (Array.isArray(filter.values)) {
         const filterVals = filter.values.map((item) => {
           // Check that filter fields aren't empty.
-          if (item.length > 0) {
+          if (item.length > 0 || typeof item === 'boolean') {
             return item
           }
         })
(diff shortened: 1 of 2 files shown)


packages/ripple-nuxt-tide/modules/search/lib/search/services/elasticsearch/lib/querybuilder.js
```diff
@@ -38,8 +38,9 @@ const addFilter = function (esbResult, filter, filterName) {
     case 'term':
       if (Array.isArray(filter.values)) {
         const filterVals = filter.values.map((item) => {
-          // Check that filter fields aren't empty.
-          if (item.length > 0) {
+          // Check that filter fields aren't empty string.
+          // If it's a boolean value, should be ignored in the check.
+          if (item.length > 0 || typeof item === 'boolean') {
```
sdpdeploy: Just to make sure I'm understanding the need for this check @tim-yao this is necessary because even when an array is passed in if the filter is a boolean and the item value is `FALSE` the check of length will return false?

GROwen: Just to make sure I'm understanding the need for this check @tim-yao this is necessary because even when an array is passed in if the filter is a boolean and the item value is FALSE the check of length will return false?

tim-yao: @GROwen yes, this condition will return any boolean value and non empty string value now. But if the value is a number, it still will fail to return. But I don't see any case and don't feel I need to fix it now.

PR #646, opened 2020-03-20
[SDPA-3748] Search results table and layout

dylankelly:

### Changed

1.  Adds search results table component `RplSearchTable` 
1. Deprecates current RplSearchResults
1. Adds RplSearchResultsV2 with slot interface for adding components - no longer supports passing component as prop.
1. Adds stories for both components

### Screenshots
localhost_3000_test-complex-listings_page=2


packages/components/Molecules/Search/searchtable.stories.js
```diff
@@ -0,0 +1,319 @@
+import RplSearchResultsTable from './SearchResultsTable'
+import { withKnobs, object } from '@storybook/addon-knobs/vue'
+
+export default {
+  title: 'Molecules/Search/RplSearchResultsTable',
```
tim-yao: May just use `SearchResultsTable` instead of `RplSearchResultsTable ` to make it same as other storybook menu items.

packages/components/Molecules/Search/index.js
```diff
@@ -1,7 +1,11 @@
 import RplSearchForm from './SearchForm.vue'
 import RplSearchResults from './SearchResults.vue'
+import RplSearchResultsTable from './SearchResultsTable.vue'
+import RplSearchResultsV2 from './SearchResultsLayout.vue'
```
tim-yao: We may add a deprecate message in `SearchResults ` to remind users to use `SearchResultsLayout `. An example: https://github.com/dpc-sdp/ripple/blob/3358427e782379eae15f500b0a8aaa34a30169f1/packages/components/Molecules/Card/CardHonourRoll.vue#L41

examples/vic-gov-au/pages/test-complex-listings.vue
```diff
@@ -0,0 +1,263 @@
...
+        :errorMsg="errorMsg"
+        :noResultsMsg="noResultsCopy"
+        :perPage="20"
+        @pager-change="changed"
+        v-if="!loading"
+      >
+        <template v-slot:results="resultsProps">
+          <rpl-col cols="full">
+            <rpl-search-results-table v-if="columns" ref="search-results-table" :items="resultsProps.searchResults" :columnConfig="columns" />
+          </rpl-col>
```
tim-yao: I think it should work with other result listing like cards by putting a for loop with some other logic here? Are you able to provide some storybook test to demo it? We need to prove it in POC that this V2 works for existing features.

examples/vic-gov-au/pages/test-complex-listings.vue
```diff
@@ -0,0 +1,263 @@
...
+  },
+  mixins: [searchMixin],
+  async asyncData ({ app, route, store }) {
+    const tideSearch = getSearch(app)
+    const fieldProfileCategoryValues = await tideSearch.setFilterOptions({
+      fieldName: 'field_vada_category_name'
+    })
+    const statuses = await app.$axios
+      .$post(
+        `https://a83890f7a31dea14e1ae83c6f0afacca.elastic.sdp.vic.gov.au/elasticsearch_index_drupal_node/_search`,
```
tim-yao: So I guess this is where I got the progress bar data. Right?

packages/components/Molecules/Search/searchtable.stories.js
dylankelly: changed

packages/components/Molecules/Search/index.js
dylankelly: Add deprecate message

examples/vic-gov-au/pages/test-complex-listings.vue
dylankelly: Added more stories including card example

examples/vic-gov-au/pages/test-complex-listings.vue
dylankelly: Yep - will need to get the url from $tideSearchConfig though.

packages/components/Molecules/Search/SearchResultsV2.vue
```diff
@@ -0,0 +1,181 @@
...
+            Displaying {{ range }} of {{ count }} results
+          </slot>
+        </div>
+        <div v-if="!!$slots.sort" class="rpl-search-results__sort">
+          <slot name="sort"></slot>
+        </div>
+      </div>
+      <template v-if="loading">
+        <slot name="loading">
+          ...loading
```
dylankelly: TODO: Add a better default loading experience. Would like design input though.

packages/components/Molecules/Search/SearchResultsV2.vue
```diff
@@ -112,7 +112,7 @@ export default {
   @import "~@dpc-sdp/ripple-global/scss/settings";
   @import "~@dpc-sdp/ripple-global/scss/tools";
 
-  .rpl-search-results-layout {
+  .rpl-search-results {
```
tim-yao: As css here is not scoped(I think same as original searchResults component), same class name here may effect on each others component? For example, user visit old search page and then v2 search page, both css will be loaded and old style will be applied on v2 page? May worth to double check and I still prefer to use different class name here in case it happened.

packages/components/Molecules/Search/SearchResultsV2.vue
tim-yao: May need some UX research on this

packages/components/Molecules/Search/searchlayout.stories.js
```diff
@@ -0,0 +1,187 @@
...
+    <rpl-search-results
+    :searchResults="searchResults"
+    :pager="pagination"
+    :count="20"
+    :perPage="10"
+  >
+    <template v-slot:results="scoped">
+      <rpl-col cols="full" v-for="(result, i) in scoped.searchResults" :key="i + '-result'">
+        <rpl-search-result
+          :title="result.title"
```
tim-yao: Looks good!
lg

packages/components/Molecules/Search/SearchResultsV2.vue
dylankelly: Ah yes good point, I might go back to calling it searchresultslayout then I was hoping it to be a drop in replacement

PR #651, opened 2020-03-31
[SDPA-3732] modern mode

tim-yao:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-3732

### Changed

1.  Changed to modern mode https://nuxtjs.org/api/configuration-modern


tim-yao:
Need add this change to upgrade script as well.

tim-yao:
Hey @dylankelly Please CR for the lighthouse test part. As the performance score is not stable, got different score every time, I would like to only test in Ripple reference site at this stage. To see how it works in the next sprint PRs.

examples/vic-gov-au/test/e2e/integration/smoke/DemoLandingPage.feature
```diff
@@ -37,6 +37,15 @@ Feature: Demo Page
     And the whats next component should exist
     And the share this component should exist
```
dylankelly, line 39: I'd recommend not running this on the DemoLanding page. Firstly its not repeatable content - the content is variable on each install and will give you varying responses. Secondly its an artificially large page, I think we should benchmark by building a landing page fixture that reflects a real world large page.

PR #658, opened 2020-04-08
Accessibility: HTML validation

alan-cole:
**JIRA issue:** https://salsadigital.atlassian.net/browse/WSM-190

### Changed

* Remove empty target="" attributes on links.
* Remove empty `<section>` generated from unused alert slot.
* Remove mask-type from `<mask>` - not allowed.
* Remove `</img>` on CampaignSecondary, due to incorrectly used v-else-if.
* Remove invalid childColsBp="[object Object]", due to blank childCols object.
* Fix invalid margin css in keydates.
* Change `<section>` to `<div>` for above / below content - avoid headerless sections.

### Screenshots

Screen Shot 2020-04-08 at 10 11 21 am


packages/components/Molecules/Layout/BaseLayout.vue
```diff
@@ -1,6 +1,5 @@
 <template>
   <div class="rpl-site-layout">
-    <section><slot name="alert"></slot></section>
```
dylankelly, line 3: I believe removing this will mean alerts wont render! Maybe the section needs to be moved to `rpl-alert` or render the section tag only if there is `$slots.alert` content. The section tag is important so the alert is an ARIA landmark.

packages/ripple-nuxt-tide/lib/core/component-loader.js
```diff
@@ -136,7 +136,7 @@ const getCols = (cols, sidebar = false) => {
     return cols
   }
 
-  return {}
+  return null
```
dylankelly, line 139: 👍 this has been annoying me for ages

packages/components/Molecules/Layout/BaseLayout.vue
alan-cole: The alerts is currently placed in the header slot, and I couldn't find anything using this slot. As discussed I will add the slot back in, but remove the surrounding section tag to fix the validation issue. Alerts can be handled separately.

PR #664, opened 2020-04-16
Set rplOptions.externalLinksInNewWindow to nuxt config value for links.

alan-cole:
**JIRA issue:** No ticket

### Changed

RplLink currently supports the ability to have any external links open in a new window.
This should be possible in nuxt.config.js with:
```
ripple: {
  externalLinksInNewWindow: true
}
```
This implementation seems to have not been fully implemented, as setting a value in nuxt.config.js does not get passed to ripple. This PR properly passes the value along to the Vue rplOptions mixin so the links correctly set a "_blank" target if external.


alan-cole:
This was implemented and QA'd as a patch on an internal project at Salsa Digital.

PR #667, opened 2020-04-17
Accessibility: Image gallery icons

alan-cole:
**JIRA issue:** https://salsadigital.atlassian.net/browse/WSM-186

### Changed

* Add accessible image gallery fullscreen icon.
* Fix storybook knobs for icon library.

Updates:
* Removed image gallery specific arrows - use default arrows.
* Implemented a mixin to include strokes on svgs.
* Fixed the youtube icon - colour is changeable, filename is correct.
* Update arrow icons - to avoid clipping on stroke, a 1px space was added to arrow icons.

### Screenshots

Screen Shot 2020-04-17 at 3 13 10 pm


packages/components/Atoms/Icon/stories.js
```diff
@@ -63,12 +70,14 @@ storiesOf('Atoms/Icon', module)
           'help',
           'home',
           'ics',
+          'image_gallery_arrow_left',
+          'image_gallery_arrow_right',
```
tim-yao, line 74: Hey @alan-cole , I feel it can be fixed by css stoke, how do you think? Instead of add another svg icon just for adding border. See https://css-tricks.com/almanac/properties/s/stroke/
It may can be a scss mixin, so we can reuse it for any svg icons.

alan-cole: I've implemented these changes, if you could re-check?

packages/components/Atoms/Global/scss/tools/_svg.scss
```diff
@@ -0,0 +1,13 @@
+@mixin rpl_svg_stroke($color, $size, $linecap: false, $dasharray: false, $dashoffset: false) {
```
tim-yao, line 1: Can we have a comment to explain how and when to use this?
Also, I feel we need a default value for color and size.

alan-cole: Added a default (black and 1px) with comment.

PR #676, opened 2020-05-07
E2E - Update social links to use button class.

alan-cole:
**JIRA issue:** No ticket.

### Changed

1.  Changed `.rpl-share-this__social` to `.rpl-share-this__button` as social class no longer exists.

NOTE: I realize this may cause incompatibilities with sites less than 1.8.0. Is that an issue?


tim-yao:
Hey @alan-cole I would not suggest to remove/rename any existing class name. Unless we 100% sure it’s safe.
I had a search in our github org, looks like Police Museum is the only place using this removed class. But after upgrade to 1.8.0, Museum custom css will stop work.

alan-cole:
I've added a new PR to address the above issue. https://github.com/dpc-sdp/ripple/pull/677

PR #679, opened 2020-05-07
[SDPA-4204] Fix publication header bug

dylankelly:

### Changed

1.  Add nested object checking in publication head meta when head not present


dylankelly:
Your eager @tim-yao !

PR #681, opened 2020-05-13
[SDPA-4224] Remove section-io-id from proxied req

dylankelly:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-4224

### Changed

1.  Remove `section-io-id` from proxy request to backend


kurtfoster:
@steveworley, I was wondering if you're looking at something here on the section side to limit headers sent to Origin? https://digital-engagement.atlassian.net/browse/SDPSUP-1606

dylankelly:
> @dylankelly , I don't think this makes sense in Ripple, it's too bespoke to our CDN. Take a look at this, https://digital-engagement.atlassian.net/browse/DDIDO-286

Agreed - Happy to remove this but the issue in the linked ticket is different to this. This addresses the issue that was identified on vicpol where simply adding the `section-io-id` to a request to BE via section causes `508 loop detected` It has nothing to do with the size of the header. Please confirm with Steve what you want to do.

PR #688, opened 2020-05-21
[SDPA-4031] Add configurable button for profile listing page

dylankelly:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-4031

### Changed

1. Move publication pagination to RplPrevNext in @dpc-sdp/ripple-pagination
1. Add deprecation notice to RplPublicationPagination
1. Add helper in $tide service to get module config from nuxt.config.js
1. Add link to honour roll listing page (configurable in nuxt.config)
1. Make route to profile listing page configurable in nuxt.config

### Screenshots
app-ripple-pr-688 lagoon vicsdp amazee io_janice-rowley


examples/vic-gov-au/nuxt.config.js
```diff
@@ -145,6 +145,10 @@ export default {
       alert: 1,
       gtm: 1
     },
+    profile: {
```
dylankelly, line 148: I would like to move this object into the modules key - I think you should be able to pass a config object in (instead of just a 1 or 0) and we should just check for truthiness

packages/components/Molecules/Pagination/PrevNext.vue
```diff
@@ -0,0 +1,159 @@
+<template>
```
dylankelly, line 1: This component is unchanged from RplPublicationPagination - just a rename

examples/vic-gov-au/nuxt.config.js
tim-yao: Agree. I am happy for you to do it in this PR or leave it in another ticket if it's too big to include in this PR(We then need to migrate this config later).

packages/components/Organisms/Publication/PublicationPagination.vue
```diff
@@ -31,9 +31,11 @@
 <script>
 import RplIcon from '@dpc-sdp/ripple-icon'
 import { RplLink, RplTextLabel } from '@dpc-sdp/ripple-link'
+import deprecate from '@dpc-sdp/ripple-global/mixins/deprecate'
```
tim-yao, line 34: Do we need to replace publication page template to use the new one? So we can remove the deprecated one in the future?

Issue #706, opened 2020-06-25
Axios POST - remove base url variable.


alan-cole:
For post requests in tide, we are setting the axios config `baseUrl`.

https://github.com/dpc-sdp/ripple/blob/master/packages/ripple-nuxt-tide/lib/core/tide.js#L80

Based on the [axios readme](https://www.npmjs.com/package/axios), `baseURL` is the correct variable name, so the `baseUrl` variable is not being properly set.
However, if the correct variable name was used and the base was being set to `config.baseUrl`, it would break form submissions, as submitting directly to the base URL from a browser would result in a CORS error.

The line should be removed as it's redundant.

tim-yao:
Thanks @alan-cole. You are right. A PR is welcome, otherwise I will do it.

alan-cole:
Thanks @tim-yao - Raised a PR.

https://github.com/dpc-sdp/ripple/pull/707

simesy:
> However, if the correct variable name was used and the base was being set to config.baseUrl, it would break form submissions, as submitting directly to the base URL from a browser would result in a CORS error.

We want to be able to configure the backend endpoint, and understand the CORS implications, so please fix it rather than removing it.

PR #707, opened 2020-06-25
Remove unused variable from tide post.

alan-cole:
**Issue:** https://github.com/dpc-sdp/ripple/issues/706

### Changed

1.  Remove setting axios variable `baseUrl`


simesy:
Don't remove this, I think this is the only way we have to point axios at an arbitrary location.  Please fix it so that `CONTENT_API_SERVER` in `.env` actually works.

The fix is `baseURL: config.baseUrl`.

simesy:
Edit - i think i'm wrong about this being the cause of my problems, since i got my setup working.

PR #710, opened 2020-06-26
[SDPA-3942] search text

tim-yao:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-3942

### Changed

1.  Changed site search text from "search website" to "search".


packages/create-ripple-app/template/_tests/_modules/test/e2e/integration/modules/search/sitesearch.feature
```diff
@@ -13,7 +13,7 @@ Feature: Search (Site search)
     Given I am using a "macbook-15" device
     When I visit the page "/"
     Then the search button in the site header should exist
-    Then the search button in the site header should have the text "Search website"
+    Then the search button in the site header should have the text "Search"
```
tim-yao, line 16: Hey @dylankelly I think I need to update these two tests in next upgrade for all sites. I think just copy that two should work. Any thoughts?

dylankelly: Sure. FYI Some sites have different text here - eg: legislation reads : "Search legislation", don't change that.

PR #731, opened 2020-07-21
[SDPA-4466] Added new var to capture robots meta from page header.

MdNadimHossain:
**JIRA issue:** Parent ticket - https://digital-engagement.atlassian.net/browse/SDPA-4457
                         Subtask - https://digital-engagement.atlassian.net/browse/SDPA-4466

### Issue
The grant nodes that get created during the feed, should not be indexed by google and also shouldn't get added to the sitemap.

### Changed

1. Added robotsMeta in the page-head.js to capture the robot's meta content from BE.
2. Added robotsMeta in tide-helper.js to render it the page header if it exists.

### Related PRs
FE PR Env link - ([Click Here](https://app-ripple-pr-731.lagoon.vicsdp.amazee.io/))
BE PR - https://github.com/dpc-sdp/tide_grant/pull/5
Content-vic PR for testing - https://github.com/dpc-sdp/content-vic-gov-au/pull/908 ([Env Link](https://nginx-php-content-vic-pr-908.lagoon.vicsdp.amazee.io/))

### Screenshots
When there is robot content - 
Screen Shot 2020-07-21 at 7 34 29 pm

When there is no robot content - 
Screen Shot 2020-07-21 at 7 35 30 pm

### DIFF
--- a/packages/ripple-nuxt-tide/lib/core/tide-helper.js
+++ b/packages/ripple-nuxt-tide/lib/core/tide-helper.js
@@ -128,7 +128,7 @@ export const getPageHeadConfig = ({
       { name: 'sitesection', content: siteSectionName },
       { name: 'content-type', content: pageType && pageType.replace('node--', '') },
       // Robots meta
-      robotsMeta ? { name: 'robots', content: robotsMeta } : ''
+      { name: 'robots', content: robotsMeta }
     ]
   }
 }
--- a/packages/ripple-nuxt-tide/lib/core/tide-helper.js
+++ b/packages/ripple-nuxt-tide/lib/core/tide-helper.js
@@ -128,7 +128,7 @@ export const getPageHeadConfig = ({
       { name: 'sitesection', content: siteSectionName },
       { name: 'content-type', content: pageType && pageType.replace('node--', '') },
       // Robots meta
-      { name: 'robots', content: robotsMeta }
+      robotsMeta ? { name: 'robots', content: robotsMeta } : ''
     ]
   }
 }
--- a/packages/ripple-nuxt-tide/lib/core/tide-helper.js
+++ b/packages/ripple-nuxt-tide/lib/core/tide-helper.js
@@ -90,7 +90,7 @@ export const stringToClass = (str) => {
  * @param {string} pageHead.imageAlt - Social sharing image alt text.
  * @param {string} pageHead.siteSectionName - Page site section name.
  * @param {string} pageHead.pageType - Page content type.
- * @param {string} pageHead.robotsNoIndex - Robots meta type.
+ * @param {string} pageHead.robotsMeta - Robots meta type.
  * @return {Object} pageHead, can be used in store tide/setPageHead.
  */
 export const getPageHeadConfig = ({
@@ -102,7 +102,7 @@ export const getPageHeadConfig = ({
   imageAlt,
   siteSectionName = '',
   pageType,
-  robotsNoIndex
+  robotsMeta
 }) => {
   return {
     htmlAttrs: {
@@ -128,7 +128,7 @@ export const getPageHeadConfig = ({
       { name: 'sitesection', content: siteSectionName },
       { name: 'content-type', content: pageType && pageType.replace('node--', '') },
       // Robots meta
-      robotsNoIndex ? { name: 'robots', content: robotsNoIndex } : ''
+      robotsMeta ? { name: 'robots', content: robotsMeta } : ''
     ]
   }
 }
--- a/packages/ripple-nuxt-tide/lib/core/tide-helper.js
+++ b/packages/ripple-nuxt-tide/lib/core/tide-helper.js
@@ -104,7 +104,7 @@ export const getPageHeadConfig = ({
   pageType,
   robotsMeta
 }) => {
-  return {
+  const head = {
     htmlAttrs: {
       lang: langcode
     },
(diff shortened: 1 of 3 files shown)


packages/ripple-nuxt-tide/lib/middleware/page-head.js
```diff
@@ -30,7 +30,8 @@ const tidePageHead = (context, pageData) => {
       siteSectionName: siteSection ? siteSection.name : '',
       image,
       imageAlt,
-      pageType: pageData.tidePage.type
+      pageType: pageData.tidePage.type,
+      robotsNoIndex: pageData.tidePage.appMetatag.robots
     }
 
     context.store.dispatch('tide/setPageHead', getPageHeadConfig(headData))
```
rubyllamadora, line 37: I can see this is being used in `tide.middleware.js`, do we need to update it in there or it's not used?

packages/ripple-nuxt-tide/lib/core/tide-helper.js
```diff
@@ -124,7 +126,9 @@ export const getPageHeadConfig = ({
       { name: 'twitter:image:alt', hid: 'hid:image:alt', content: imageAlt },
       // Custom page meta
       { name: 'sitesection', content: siteSectionName },
-      { name: 'content-type', content: pageType && pageType.replace('node--', '') }
+      { name: 'content-type', content: pageType && pageType.replace('node--', '') },
+      // Robots meta
+      robotsNoIndex ? { name: 'robots', content: robotsNoIndex } : ''
```
rubyllamadora, line 131: 👍

rubyllamadora: Do we have some unit test around this function? Sorry probably me unfamiliar with the process 😬

packages/ripple-nuxt-tide/lib/middleware/page-head.js
MdNadimHossain: Hey, @rubyllamadora I think for all the page header we are capturing the data here and if we want to handle these values differently for a pagetype then we do that in `tide.middleware.js`

packages/ripple-nuxt-tide/lib/core/tide-helper.js
MdNadimHossain: @rubyllamadora I don't think we have unit test placed for this function :disappointed:

packages/ripple-nuxt-tide/lib/core/tide-helper.js
```diff
@@ -100,7 +101,8 @@ export const getPageHeadConfig = ({
   image,
   imageAlt,
   siteSectionName = '',
-  pageType
+  pageType,
+  robotsMeta
```
tim-yao, line 105: Hey @MdNadimHossain Do you think we can pass all meta tags from page data as a property instead of specific one? So we don't need to add more properties in the future. This function was created by me in a quick code split task didn't consider too much on the function its self. And it's only used by packages/ripple-nuxt-tide/lib/middleware/page-head.js. Feel free to improve it.

packages/ripple-nuxt-tide/lib/core/tide-helper.js
tim-yao: Hey @MdNadimHossain @rubyllamadora I didn't add unit test for this one. But I should. Sorry for that. All the unit tests are in https://github.com/dpc-sdp/ripple/tree/master/packages/ripple-nuxt-tide/test/unit
And for tide-helper tests, all in https://github.com/dpc-sdp/ripple/blob/master/packages/ripple-nuxt-tide/test/unit/tide.test.js.

Can you add one for this in this PR? Cheers!

MdNadimHossain: @tim-yao @rubyllamadora I have added the test.

packages/ripple-nuxt-tide/lib/core/tide-helper.js
MdNadimHossain: Hey @tim-yao Correct me if I am wrong, but if we do that then we have to do the validation of them in the function itself. Now the validation is done for each of them in here `packages/ripple-nuxt-tide/lib/middleware/page-head.js` isn't it!

tim-yao: @MdNadimHossain yes, you are right. It's ok to keep it as it is then. No need to over engineering it.

packages/ripple-nuxt-tide/lib/core/tide-helper.js
```diff
@@ -124,7 +126,9 @@ export const getPageHeadConfig = ({
       { name: 'twitter:image:alt', hid: 'hid:image:alt', content: imageAlt },
       // Custom page meta
       { name: 'sitesection', content: siteSectionName },
-      { name: 'content-type', content: pageType && pageType.replace('node--', '') }
+      { name: 'content-type', content: pageType && pageType.replace('node--', '') },
+      // Robots meta
+      { name: 'robots', content: robotsMeta }
```
tim-yao, line 131: If no robots content, should we just not add it? @MdNadimHossain

MdNadimHossain: @tim-yao yes, the check was there but then I removed it thinking empty check at `page-head.js` will not add it, but yeah I have added the check again here like this - `robotsMeta ? { name: 'robots', content: robotsMeta } : ''`

PR #732, opened 2020-07-22
[SDPA-4133] Add Hover State to Feedback Form Cancel Button

Yuuupie:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-4133

### Changed

1.  Hover colour/state of cancel button in feedback form.

### Screenshots

image


packages/ripple-nuxt-tide/modules/webform/components/TideContentRating.vue
```diff
@@ -250,7 +250,11 @@ $app-content-text-color: rpl-color('dark_neutral') !default;
   }
 
   .app-button--cancel {
-    background-color: rpl-color('danger');
+    background-color: $rpl-button-danger-background-color;
+  }
+
+  .app-button--cancel:hover {
+    background-color: $rpl-button-quick-exit-hover-background-color;
```
tim-yao, line 257: Hey @Yuuupie can we add a new var called `$rpl-button-danger-hover-background-color` in "packages/components/Atoms/Global/scss/settings/_button.scss" ? Even it's same value, but easier for maintaining. Thanks!

Yuuupie: Yeah good idea

packages/ripple-nuxt-tide/modules/webform/components/TideContentRating.vue
```diff
@@ -250,7 +250,11 @@ $app-content-text-color: rpl-color('dark_neutral') !default;
   }
 
   .app-button--cancel {
-    background-color: rpl-color('danger');
+    background-color: $rpl-button-danger-background-color;
```
tim-yao, line 253: This is nice to see you fix it. 👍  All color in components need a meaningful var instead of directly use color.

packages/components/Atoms/Global/scss/settings/_button.scss
```diff
@@ -15,7 +15,7 @@ $rpl-button-secondary-text-color: rpl_color('white') !default;
 $rpl-button-secondary-hover-background-color: rpl_color('primary') !default;
 $rpl-button-secondary-hover-text-color: $rpl-button-secondary-text-color !default;
 $rpl-button-danger-background-color: rpl_color('danger') !default;
-$rpl-button-quick-exit-hover-background-color: lighten(rpl-color('danger'), 5%) !default;
```
tim-yao, line 18: It will be dangerous to remove an existing variable as it will cause breaking change when some customer projects use it. We have to make sure it's safe to do otherwise we'd rather keep it and deprecate it. 
However I have did a global search for this one in all projects and it's safe.

Yuuupie: Oh I see, will keep that in mind next time.

Yuuupie:
@tim-yao Is it a problem that test is failing?

tim-yao:
> @tim-yao Is it a problem that test is failing?
Demo Page -- Demo landing page loads correctly (failed)

Hey @Yuuupie in the smoke test, looks like related links are not able to be found. It looks like nothing related to your PR, but worth to look. Are you able to have a look why?

PR #733, opened 2020-07-23
[SDPA-4468] New webform validation

MdNadimHossain:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-4468

### Issue 
2. When multiple select is true, it doesn't validate the field for the required value.
1. When `webform_term_select` field has a limit of max value at the BE, the user still can select more than the max value and ends up throwing server error during form submission.

### Changed

Two Custom validation added. One will check if a field is required then at least one option should be selected. 
2nd validation will check if user selects more than the limit of a max selection.

### Future work
The automation test needs to be added for form elements. For example, demo content at the BE for a form, with all the allowed form elements and then adding test according to that at the FE.

### Screenshots
Screen Shot 2020-07-23 at 11 33 35 am
Screen Shot 2020-07-23 at 11 33 59 am


packages/components/Molecules/Form/index.vue
```diff
@@ -84,6 +84,20 @@ export default {
...
+      }
+      return ['Add a selection']
+    }
+    // Validate if multiple select has a max limiit of selection.
+    VueFormGenerator.validators.rplSelectMaxLimit = function (value, field) {
+      if (value && value.length <= field.max) {
+        return []
+      }
+      return ['More than ' + field.max + ' selections are not allowed']
+    }
```
tim-yao, line 100: @MdNadimHossain I was planning to move these custom validators out of this component and into a separated js file. To make code more easy to maintain. Happy for you to do a little refactoring if you have time. But not necessary. Thanks!

PR #736, opened 2020-07-27
[SDPA-4235] optimise svg icon component to improve performance - POC

rubyllamadora:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-4235

### Changed

1.  Refactor Icon component to improve performance

Note: I have compared 6 icons below just to see performance comparison. Tried to install component performance plugin in my local to check detailed component speed but hasn't been successful

#### Icons listing
**Current SVG** 
Screen Shot 2020-08-10 at 1 48 24 pm
**Refactored**
Screen Shot 2020-08-10 at 1 48 16 pm
 
#### Text with Icon
Screen Shot 2020-08-10 at 2 07 25 pm
Screen Shot 2020-08-10 at 2 07 31 pm


tim-yao:
@rubyllamadora this vue-awesome approach looks nice! I did a check in my local and the icon load time is less now. 👍  One more thing I would like to check in the POC is custom icons. Can you please continue have a look at it, some docs in  https://github.com/dpc-sdp/ripple/blob/master/packages/components/Atoms/Icon/stories.js#L170
There is a few sites are having custom icons, we need to continue allow customer to override the icons or register their own icons. I feel it will be a breaking change for those have custom icons.

rubyllamadora:
> @rubyllamadora this vue-awesome approach looks nice! I did a check in my local and the icon load time is less now. 👍 One more thing I would like to check in the POC is custom icons. Can you please continue have a look at it, some docs in https://github.com/dpc-sdp/ripple/blob/master/packages/ripple-nuxt-tide/README.md#custom-icons
> There is a few sites are having custom icons(e.g vic-gov-au), we need to continue allow customer to override the icons or register their own icons. I feel it will be a breaking change for those have custom icons.

Thanks @tim-yao for taking a look. Definitely breaking change :( 
Will have a look at how we can do custom icons. I'm thinking they should be able to register their own icons and add it to the arrays but the implementation from their side will be refactored. Will check and get back to this.

examples/vic-gov-au/assets/ripple-icon/icon_circle_arrow_right.js
```diff
@@ -0,0 +1,13 @@
...
+  'icon_circle_arrow_right': {
+    width: 10,
+    height: 10,
+    paths: [
+      {
+        d: 'M0 10c0 5.524 4.476 10 10 10s10-4.476 10-10S15.524 0 10 0 0 4.476 0 10zm18.065 0A8.062 8.062 0 0110 18.065 8.062 8.062 0 011.935 10 8.062 8.062 0 0110 1.935 8.062 8.062 0 0118.065 10zM10.44 4.762l4.895 4.895c.19.19.19.496 0 .686l-4.895 4.895a.484.484 0 01-.686 0l-.79-.79a.483.483 0 01.008-.694l2.834-2.706H5.323a.485.485 0 01-.484-.483v-1.13c0-.266.217-.483.484-.483h6.483L8.972 6.246a.483.483 0 01-.008-.694l.79-.79c.19-.19.496-.19.686 0z'
+      }
+    ]
+  }
+})
```
rubyllamadora, line 13: This will be how to register custom icons/if they want to replace Ripple icon with their own they just have to register it with the same name as in Ripple.

examples/vic-gov-au/assets/ripple-icon/icon_circle_arrow_right.js
```diff
@@ -0,0 +1,13 @@
+import Icon from '../../../../packages/components/Atoms/Icon/Icon'
```
rubyllamadora, line 1: In e.g. vic-gov-au, the import path will be '@dpc-sdp/ripple-icon'. I just use relative path here for example

examples/vic-gov-au/pages/CustomIcon.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+  <rpl-icon symbol="icon_circle_arrow_right" color="primary" size="m" />
+</template>
+
+<script>
+import RplIcon from '../../../packages/components/Atoms/Icon/Icon'
+import './../assets/ripple-icon/icon_circle_arrow_right'
```
rubyllamadora, line 7: This import here is for icons specific to their website and should not be added in Ripple. Or if they want to override Ripple icon they just register the Icon as the same name in Ripple.

This component is an example of how to implement custom icons and what needs to be refactored if we want to go down this path. 

ping @tim-yao
Screen Shot 2020-08-13 at 2 26 58 pm

rubyllamadora: The import can also be added inside ripple: {} config when `customIcon:true`. We just need index.js with Icon registration inside ripple-icon directory so we can call it in e.g.
```
<% if (options.customIcon) { %>
  // Add custom icons to library.
  import '../assets/ripple-icon/'
<% } %>
```

rubyllamadora:
Closing and will revive changes when ticket is ready to be worked on.

PR #738, opened 2020-07-29
Feature/globalise border radius variable

rubyllamadora:

### Changed

1.  Update border-radius in components to retrieve from global variable
2.  Embed IE11 polyfill

### DIFF
--- a/packages/components/Molecules/Breadcrumbs/index.vue
+++ b/packages/components/Molecules/Breadcrumbs/index.vue
@@ -34,7 +34,7 @@ export default {
 
   $rpl-breadcrumbs-background: rpl-color('white') !default;
   $rpl-breadcrumbs-border: 1px solid rpl-color('mid_neutral_1') !default;
-  $rpl-breadcrumbs-border-radius: var(--rpl-theme-border-radius, rem(4px)) !default;
+  $rpl-breadcrumbs-border-radius: $rpl-border-radius !default;
   $rpl-breadcrumbs-padding: ($rpl-space-3) ($rpl-space * 5) !default;
   $rpl-breadcrumbs-link-ruleset: ('xs', 1.4em, 'medium') !default;
   $rpl-breadcrumbs-link-color: rpl_color('primary') !default;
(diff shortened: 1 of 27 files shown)


packages/ripple-nuxt-tide/lib/layouts/default.vue
```diff
@@ -124,6 +124,13 @@ export default {
       this.anchorScrollFix(this.$route.hash)
     }
     RplLinkEventBus.$on('navigate', this.onNavigate)
+
+    if (window.MSInputMethodContext && document.documentMode) {
```
rubyllamadora, line 128: @dylankelly I put it here since `window` property does not exist in nuxt.config.js as it server renders. Otherwise, happy to move this to config so this script loads regardless of browser type.

dylankelly: @rubyllamadora You can check which environment you are in by using `process.client` or `process.server` https://nuxtjs.org/faq/window-document-undefined . So in this case a ternary check for process.client should work. Can you please move this out of the unrelated PR for border radius? I'd like to merge them separately, and its always a good practice to keep PR's only about one thing.

rubyllamadora: I have reverted the change. I'll create new PR for this 👍

packages/components/Molecules/Breadcrumbs/index.vue
```diff
@@ -34,7 +34,7 @@ export default {
 
   $rpl-breadcrumbs-background: rpl-color('white') !default;
   $rpl-breadcrumbs-border: 1px solid rpl-color('mid_neutral_1') !default;
-  $rpl-breadcrumbs-border-radius: rem(4px) !default;
+  $rpl-breadcrumbs-border-radius: var(--rpl-theme-border-radius, rem(4px)) !default;
```
dylankelly, line 37: This might be better served by a SCSS variable - What if you wanted to change the default value everywhere? If you created a variable `$rpl-theme-border-radius: var(--rpl-theme-border-radius, rem(4px))` you would only have one place to change it.

rubyllamadora: Ahhh true. I have thought about that when I asked you about the flexibility. I'll create a global variable. Thanks

PR #746, opened 2020-08-10
[SW-288] : Fix filter disappearing after no results

FleetAdmiralButter:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SW-288

### Changed

1.  We always include filter aggs in the query to ElasticSearch.
2. Filter options are always refreshed on the search form because ElasticSearch will return an aggregation prop with the response.

### DIFF
--- a/packages/ripple-nuxt-tide/modules/search/lib/search/services/elasticsearch/lib/api.js
+++ b/packages/ripple-nuxt-tide/modules/search/lib/search/services/elasticsearch/lib/api.js
@@ -31,7 +31,7 @@ export default ({
       // Only update the filter options if a selection hasn't already been made, otherwise the filter would need to be
       // cleared before another selection can be made.
       if (!filters[inputField.model]) {
-        if (inputField.type === 'rplchecklist' && inputField.values.length > 0 && (typeof inputField.filter.computedFilter === 'undefined' || inputField.filter.computedFilter === false)) {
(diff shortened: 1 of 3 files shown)


packages/ripple-nuxt-tide/modules/search/lib/search/services/elasticsearch/lib/api.js
```diff
@@ -31,7 +31,7 @@ export default ({
       // Only update the filter options if a selection hasn't already been made, otherwise the filter would need to be
       // cleared before another selection can be made.
       if (!filters[inputField.model]) {
-        if (inputField.type === 'rplchecklist' && inputField.values.length > 0 && (typeof inputField.filter.computedFilter === 'undefined' || inputField.filter.computedFilter === false)) {
+        if (inputField.type === 'rplchecklist' && (typeof inputField.filter.computedFilter === 'undefined' || inputField.filter.computedFilter === false)) {
```
GROwen, line 34: This could certainly fix the issue but I'm not sure it won't introduce another.

The check of the the length was to ensure that a filter value existed before adding the aggregation. This does sound like it would cause an issue when used to simply populate the filters.

For that reason I think a refactor of how the filters are populated needs to happen so it's more logically separated from a query fetching filtered results.

If I perform a successful search on the PR env then add a filter I see results returned (in the dev-tools console) but the application throws an error.

FleetAdmiralButter:
Needs a rethink, extra debugging pushed to https://github.com/dpc-sdp/premier-vic-gov-au/pull/86/files

PR #753, opened 2020-08-19
[SDPA-4501] add extra validation to search result mapping

rubyllamadora:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-4501

### Changed

1.  Updated validation in search result mapping to ensure results are being displayed with no issues


packages/components/Molecules/Search/__tests__/RplSearchForm.test.js
```diff
@@ -0,0 +1,78 @@
...
+    expect(wrapper.vm.filterCount).toEqual(2)
+  })
+
+  it('tests watcher for prefillSearchTerm to assign value to searchInput correctly', () => {
+    const wrapper = shallowMount(SearchForm)
+
+    wrapper.setProps({ prefillSearchTerm: 'something' })
+    expect(wrapper.vm.searchInput).toEqual('something')
+  })
+})
```
rubyllamadora, line 78: I was a bit fiddly and added this. Not directly related to ticket but ran into this when I was debugging search.

packages/ripple-nuxt-tide/modules/search/lib/searchmixin.js
```diff
@@ -80,6 +81,10 @@ const searchMixin = {
       } catch (e) {
         const msg = 'Search isn\'t working right now, please try again later.'
         this.errorMsg = msg
+
+        if (process.server) {
+          logger.error('Retrieving search results failed in getSearchResults()', { e, label: 'SearchMixin' })
```
tim-yao, line 86: Hey @rubyllamadora Thanks for adding the log here. The error object need to be added as `{ error: e, label: 'SearchMixin'}` otherwise it won't be used.

packages/components/Molecules/Search/__tests__/RplSearchForm.test.js
tim-yao: > I was a bit fiddly and added this. Not directly related to ticket but ran into this when I was debugging search.

That looks nice to me. 👍  No need to cover all logics in this ticket. That will require a lot effort and background knowledge. We can add tests for any search bug fixes later.

tim-yao (approved):

packages/ripple-nuxt-tide/modules/search/lib/searchmixin.js
rubyllamadora: Nice pickup! I'll rename parameter to error so can just have `{error, label:...` - cleaner :P

PR #771, opened 2020-09-09
[DDIDO-186] Added non-prod and Fixed Prod Routes

Yuuupie:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/DDIDO-186

### Changed

1.  Added routes for master and develop.
2.  Made prod route go under prod environment instead of master.


tim-yao (approved):

.lagoon.yml
```diff
@@ -11,16 +11,28 @@ tasks:
...
+  production:
     routes:
       - storybook:
         - "www.ripple.sdp.vic.gov.au":
             tls-acme: 'false'
             insecure: Allow
             hsts: max-age=31536000
-        - "ripple.sdp.vic.gov.au":
+    monitoring_urls:
+      - "www.ripple.sdp.vic.gov.au"
```
FleetAdmiralButter, line 22: Do we need to have monitoring set up for storybook?

FleetAdmiralButter (approved):
Looks like Lagoon successfully parsed this in the build.

.lagoon.yml
Yuuupie: Well storybook is the service that the routes point to so yes, but I'm guessing you might mean do we need to monitor the app route then? For that I don't know.

FleetAdmiralButter: Sorry - I meant as in since Storybook isn't really a consumer facing site.

Yuuupie: I get you now. @kurtfoster Any thoughts?

kurtfoster: @Yuuupie / @FleetAdmiralButter , I don't think we need to monitor storybook.

tim-yao (approved):

kurtfoster (approved):

PR #772, opened 2020-09-09
[SDPSUP-2024] Add html lang attr from meta tag

dylankelly:

### Changed

1.  Adds html lang attribute based on Drupal `Content Language` meta. It seems langcode is not populated in JSON api from this value

### DIFF
--- a/packages/ripple-nuxt-tide/lib/core/tide-helper.js
+++ b/packages/ripple-nuxt-tide/lib/core/tide-helper.js
@@ -46,8 +46,14 @@ export const metatagConverter = (metatagNormalized) => {
     switch (element.tag) {
       case 'meta':
         if (element.attributes['http-equiv']) {
-          metatags.lang = element.attributes.content
-        } else {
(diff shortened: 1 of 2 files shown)


packages/ripple-nuxt-tide/lib/core/tide-helper.js
```diff
@@ -45,7 +45,11 @@ export const metatagConverter = (metatagNormalized) => {
   metatagNormalized.forEach(element => {
     switch (element.tag) {
       case 'meta':
-        metatags[element.attributes.name] = element.attributes.content
+        if (element.attributes['http-equiv']) {
+          metatags.lang = element.attributes.content
```
tim-yao, line 49: Hey @dylankelly this `http-equiv` can be something else like below. So we need a condition to check if it's `content-language`.
Screen Shot 2020-09-09 at 5 22 15 pm

You can leave it to me, I will add this one and do a QA. Cheers.

PR #782, opened 2020-09-18
[SDPA-2995] Preview links with authentication

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2995

### Changed

**Note: This is a work in progress.**

*  Uses nuxtjs auth module.

### Screenshots

Preview Link_ Login
Preview Link_ Log out


alan-cole:
Moved to PR: https://github.com/dpc-sdp/ripple/pull/793

PR #791, opened 2020-09-23
Update callout background colour

rubyllamadora:

### Changed

1.

### DIFF
--- a/packages/components/Atoms/Global/scss/settings/_colorcore.scss
+++ b/packages/components/Atoms/Global/scss/settings/_colorcore.scss
@@ -3,7 +3,6 @@ $rpl-colors-core: (
   'primary': #0052c2,
   'dark_primary': #023b89,
   'secondary': #0095ec,
-  'light_secondary': #e6f4fd,
   'extra_dark_neutral': #011a3c,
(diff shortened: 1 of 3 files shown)


tim-yao (approved):
I was thinking we should have a global background color set to be used by different component instead of design for each component. So it make manage these color more easily.

rubyllamadora:
> I was thinking we should have a global background color set to be used by different component instead of design for each component. So it make manage these color more easily.

👍 this requires update in Ripple dependent websites? I feel like each of our colours e.g. primary and secondary should have lighter and darker equivalent. We already have dark_primary but no light or lighter.

tim-yao:
> > I was thinking we should have a global background color set to be used by different component instead of design for each component. So it make manage these color more easily.
> 
> 👍 this requires update in Ripple dependent websites? I feel like each of our colours e.g. primary and secondary should have lighter and darker equivalent. We already have dark_primary but no light or lighter.

Yes, it was following the Ripple design, it does need to be reworked to including full set of lighter and darker equivalent for both primary and secondary color.

tim-yao (approved):

PR #793, opened 2020-09-24
[SDPA-2995] Preview links with authentication

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-2995

### Changed

**Note: This is a work in progress.**

*  Uses nuxtjs auth module.

### Screenshots

Preview Link_ Login
Preview Link_ Log out


packages/ripple-nuxt-tide/lib/core/tide.js
```diff
@@ -420,19 +420,18 @@ export const tide = (axios, site, config) => ({
     return pageData
   },
 
-  getPreviewPage: async function (contentType, uuid, revisionId, section, params, headersConfig) {
-    if (revisionId === 'latest') {
-      params.resourceVersion = 'rel:working-copy'
-    } else {
-      params.resourceVersion = `id:${revisionId}`
-    }
+  getPageByPreviewLink: async function (path, section) {
```
tim-yao, line 423: Hey @alan-cole should we move this into preview module packages/ripple-nuxt-tide/modules/preview/lib/helpers.js ?

packages/ripple-nuxt-tide/modules/preview/module.js
```diff
@@ -0,0 +1,56 @@
...
+          name: 'X-OAuth2-Authorization'
+        },
+        refreshToken: {
+          property: 'refresh_token',
+          maxAge: 1209600
+        },
+        responseType: 'code',
+        grantType: 'authorization_code',
+        clientId: options.oauth.clientId,
+        scope: ['editor', 'authenticated']
```
tim-yao, line 37: @alan-cole If you add option `resetonerror: true` here(https://dev.auth.nuxtjs.org/api/options#resetonerror), user will be automatically logged out if an error happens(For example when token expired). Then we don't need to handle it by ourselves in axios onError event anymore.

alan-cole: Should be an easy fix. I'll look at that.

packages/ripple-nuxt-tide/lib/core/tide.js
alan-cole: Yep, will need to import jsonapiParse and tide, and export this function for use in middleware. One question, are you fine with isPreviewPath being in core, or would you like that in the preview module too?

tim-yao: @alan-cole Thanks. Let's leave `isPreviewPath ` in core. As it need to be used by core code.

packages/ripple-nuxt-tide/lib/core/tide.js
```diff
@@ -420,19 +420,18 @@ export const tide = (axios, site, config) => ({
...
+
+    params.resourceVersion = (revisionId === 'latest') ? 'rel:working-copy' : `id:${revisionId}`
 
     const pathData = {
       entity_type: 'node',
       bundle: contentType,
       uuid: uuid
     }
-    const entity = await this.getEntityByPathData(pathData, params, headersConfig)
+    const entity = await this.getEntityByPathData(pathData, params)
```
tim-yao, line 434: Can we add `headersConfig ` back? It will be used for passing extra headers like request id.

packages/ripple-nuxt-tide/modules/preview/module.js
tim-yao: @alan-cole My full fix is in https://github.com/dpc-sdp/ripple/pull/796/files

alan-cole: Cherry picked your commit into the new branch.

packages/ripple-nuxt-tide/lib/core/tide.js
alan-cole: Added

alan-cole:
Code has been merged into https://github.com/dpc-sdp/ripple/pull/802. Please review there.

PR #786, opened 2020-09-23
[SDPA-3202] Share links

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-3202

### Changed

1.  Updated middleware to check + process: share links, else preview links, else all other page requests.
2. Added processing of share links - converting the share URL into a getEntityByPathData() call.
3. Added isSharePath() to new /lib/path.js
4. Moved isPreviewPath() check out of authenticated-content module and into core /lib/path.js file. This may be a **breaking change** if other sites import isPreviewPath() from authenticated-content.
5. Updated tests.

### Screenshots

With:
`CONTENT_API_SERVER=https://nginx-php-content-vic-pr-971.lagoon.vicsdp.amazee.io/`
Going to: http://localhost:3000/share_link/adda2e43-9205-4103-8431-051ca394a0e5/9082?section=298
Will show the page:
Screen Shot 2020-09-23 at 11 31 54

PR #794, opened 2020-09-24
[SDPA-3202] Share Links - Test Environment. DO NOT MERGE.

alan-cole:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-3202

### Changed

1.  This PR is https://github.com/dpc-sdp/ripple/pull/786 but with the .env set to https://nginx-php-content-vic-pr-971.lagoon.vicsdp.amazee.io/


packages/ripple-nuxt-tide/lib/core/tide.js
```diff
@@ -443,6 +448,33 @@ export const tide = (axios, site, config) => ({
     return pageData
   },
 
+  getPageByShareLink: async function (path, section) {
```
tim-yao, line 451: @alan-cole Can we split this function to different file? I would like to make tide.js only focus on the basic Tide API communication. Thanks!

tim-yao (commented):
Hey @alan-cole I think if we turn off authenticated content module, this should still work. Right?

packages/ripple-nuxt-tide/lib/core/tide.js
alan-cole: I'll create a submodule for sharelinks and add this in.

alan-cole:
> Hey @alan-cole I think if we turn off authenticated content module, this should still work. Right?

Hi Tim, that's correct. This should still be working without authenticated content.

alan-cole:
Closing as Share links is now part of develop.

PR #799, opened 2020-10-02
[SDPA-4316] quick exit by section

tim-yao:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-4316

### Changed

1.  Updated quickexit logic to handle site section level
2. Added a global page mixin for the ability to set ripple option on every page.

### DIFF
--- a/packages/ripple-nuxt-tide/lib/core/utils/site-section.js
+++ b/packages/ripple-nuxt-tide/lib/core/utils/site-section.js
@@ -1,0 +1,21 @@
+/**
+ * Get the site section data
+ * @param {string|number} sectionId
+ * @param {Object[]} sites - Tide field_node_site data
+ * @returns {Object} site section data
+ */
+const getSiteSectionData = (sectionId, sites) => {
+  sectionId = parseInt(sectionId)
+  const site = sites.find(site => site.drupal_internal__tid === sectionId)
+  if (site) {
+    return {
+      id: sectionId,
+      name: site.name,
+      quickexit: site.field_site_show_exit_site
+    }
+  } else {
+    return null
+  }
+}
+
+export { getSiteSectionData }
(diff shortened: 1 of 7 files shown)


packages/ripple-nuxt-tide/lib/middleware/sidebar.js
```diff
@@ -65,6 +65,8 @@ const tideSideBar = async (context, pageData, headersConfig) => {
   }
   // site section
   if (pageData.tidePage.section) {
+    // TODO: SDPA-4675 It's unnecessary query site data here, as we can get the site section data from page field_node_site, which includes the menu ids.
+    // We should just get the menu here, which can reduce Tide requests number from at 3 to 1 here.
```
rubyllamadora, line 69: 👍

packages/ripple-nuxt-tide/lib/core/utils/site-section.js
```diff
@@ -0,0 +1,21 @@
...
+ * @returns {Object} site section data
+ */
+const getSiteSectionData = (sectionId, sites) => {
+  sectionId = parseInt(sectionId)
+  const site = sites.find(site => site.drupal_internal__tid === sectionId)
+  if (site) {
+    return {
+      id: sectionId,
+      name: site.name,
+      quickexit: site.field_site_show_exit_site
```
rubyllamadora, line 14: I'm assuming this will always have a value. Or do we need to add defensive or `|| false`?

rubyllamadora (approved):
Mixin looks good (: and Quick exit is displaying ok in pages @tim-yao 
I think it's the current functionality but when you click on Quick exit, does it always open a new tab at the same time redirects your current page to google too?

packages/ripple-nuxt-tide/lib/core/utils/site-section.js
tim-yao: @rubyllamadora good pick! It will be `null`, fixed so we can make sure it's a boolean.

PR #796, opened 2020-09-28
[SDPA-2995] add error handling

tim-yao:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-

### Changed

1.

PR #802, opened 2020-10-05
[SDPA-2995] [SDPA-3202] Preview and share links

alan-cole:
### Summary

This PR contains the the changes and CR remediations from:

1. https://github.com/dpc-sdp/ripple/pull/786
2. https://github.com/dpc-sdp/ripple/pull/793
3. https://github.com/dpc-sdp/ripple/pull/796

This has been placed in the 1 branch as the preview module now contains preview links and share links.

**JIRA issue:**
https://digital-engagement.atlassian.net/browse/SDPA-2995
https://digital-engagement.atlassian.net/browse/SDPA-3202

### Changed

**PREVIEW**
*  Uses nuxtjs auth module.

**SHARE LINKS**
1.  Updated middleware to check + process: share links, else preview links, else all other page requests.
2. Added processing of share links - converting the share URL into a getEntityByPathData() call.
3. Added isSharePath() to new /lib/path.js
4. Moved isPreviewPath() check out of authenticated-content module and into core /lib/path.js file. This may be a **breaking change** if other sites import isPreviewPath() from authenticated-content.
5. Updated tests.

### Screenshots

Preview Link_ Login
Preview Link_ Log out
With:
`CONTENT_API_SERVER=https://nginx-php-content-vic-pr-971.lagoon.vicsdp.amazee.io/`
Going to: http://localhost:3000/share_link/adda2e43-9205-4103-8431-051ca394a0e5/9082?section=298
Will show the page:
Screen Shot 2020-09-23 at 11 31 54


packages/ripple-nuxt-tide/modules/preview/lib/helpers.js
```diff
@@ -0,0 +1,93 @@
+import { redirect } from '../config/paths'
+import jsonapiParse from 'jsonapi-parse'
+
+async function tidePreview (context, headersConfig) {
+  if (!context.$auth.loggedIn) {
+    return context.redirect(`${redirect.login}?destination=${context.req.url}`)
+  } else {
+    return getPageByPreviewLink(context.app.$tide, context.route.path, context.store.state.tide.siteId, context.route.query.section, headersConfig)
```
tim-yao, line 8: `context.store.state.tide.siteId` is not the site section id. A page is always under a site and also optionally under a site section. We should get the section id from preview url query string. e.g `/preview/landing_page/d5457459-1735-4b80-8fc6-46e6bdfd74b3/latest?section=224`

packages/ripple-nuxt-tide/modules/preview/lib/helpers.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+  const pageData = jsonapiParse.parse(entity).data
+
+  // Append the site section to page data
+  pageData.section = (section && section !== site) ? section : null
+  return pageData
+}
+
+async function tideShare (context, headersConfig) {
+  return getPageByShareLink(context.app.$tide, context.route.path, context.store.state.tide.siteId, context.route.query.section, headersConfig)
```
tim-yao, line 58: Same site id is not section id here.

packages/ripple-nuxt-tide/modules/preview/pages/Login.vue
```diff
@@ -0,0 +1,67 @@
...
+      this.loggingOut = true
+      await this.$auth.logout()
+    }
+  },
+  async asyncData ({ app, route, store }) {
+    const title = !app.$auth.loggedIn ? 'Logging in' : 'Log in'
+    store.dispatch('tide/setPageHead', { title })
+  },
+  mounted () {
+    tidePreviewLogin(this.$route, this.$auth, localStorage)
```
tim-yao, line 64: Is there any reason we use localStorage instead of Vue store? I think you mentioned once but I couldn't remember. 🤦

packages/ripple-nuxt-tide/modules/preview/README.md
```diff
@@ -0,0 +1,24 @@
...
+{
+  tide: {
+    modules: {
+      preview: 1
+    }
+  }
+}
+```
+
+Add `CONTENT_API_CLIENT_ID` to `.env`.
```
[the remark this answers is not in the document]
alan-cole: Updated.

packages/ripple-nuxt-tide/modules/preview/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+```js
+{
+  tide: {
+    modules: {
+      preview: 1
+    }
+  }
+}
+```
```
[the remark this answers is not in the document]
alan-cole: Updated.

examples/vic-gov-au/nuxt.config.js
```diff
@@ -123,6 +123,9 @@ export default {
       username: process.env.CONTENT_API_AUTH_USER,
       password: process.env.CONTENT_API_AUTH_PASS
     },
+    oauth: {
+      clientId: process.env.CONTENT_API_CLIENT_ID
```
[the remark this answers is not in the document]
alan-cole: The reason I added it in this way was to get the test environments set up. While we could access the env variable directly in the preview/module.js, accessing it through the nuxt config does feel like the right way to do it, considering most other env variables are accessed this way. I was also wondering about potential future changes, if scopes need to expand and need to be set unique for some sites, this could be a place where those options are set as well.
But I'm happy to change if it's going to make it easier for deployment.

packages/ripple-nuxt-tide/modules/preview/module.js
```diff
@@ -0,0 +1,67 @@
...
+module.exports = function () {
+  const options = this.options.tide
+
+  this.options.proxy = {
+    ...this.options.proxy,
+    '/oauth/token': {
+      target: options.baseUrl,
+      onProxyReq (proxyReq, req, res) {
+        proxyReq.removeHeader('authorization')
+      }
```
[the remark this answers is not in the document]
alan-cole: Added.

packages/ripple-nuxt-tide/modules/preview/config/paths.js
```diff
@@ -0,0 +1,8 @@
+const redirect = {
```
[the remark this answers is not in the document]
alan-cole: Added.

packages/ripple-nuxt-tide/modules/preview/pages/Login.vue
alan-cole: Easier to set up basically. Vuex store requires more lines of code. Also, the server doesn't need to know of the redirect, and adding to the store includes more data passing than needed. Though, we do use the store for most other things, so happy to change if you need.

packages/ripple-nuxt-tide/modules/preview/lib/helpers.js
alan-cole: This getPageByPreviewLink() is based on it's previous tide.js implementation where the $tide module is passed a site id at construction, and the functions like getPageByPath() / getPageByPreviewLink() / getPageByShareLink() all perform use that value to compare to a section ID: 
`pageData.section = pathData.section === site ? null : pathData.section`

In order to move this function into it's own module, keep the same implementation and have access to a site id, I used the value stored in Vuex to pass through to the function. You can see the section id from the query string is also passed as a parameter directly after the site id.

If the `pageData.section` logic needs to change and doesn't require the site ID anymore, perhaps the implementations change for getPageByPath() / getPageByPreviewLink() / getPageByShareLink() ?

packages/ripple-nuxt-tide/modules/preview/module.js
```diff
@@ -0,0 +1,67 @@
...
+      drupal: {
+        scheme: 'oauth2',
+        endpoints: {
+          authorization: `${options.baseUrl}oauth/authorize`,
+          token: `/oauth/token`
+        },
+        token: {
+          property: 'access_token',
+          type: 'Bearer',
+          maxAge: 300,
```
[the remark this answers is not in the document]
alan-cole: I've set the following, as per the implementation on the server:
```
token: {
  property: 'access_token',
  type: 'Bearer',
  maxAge: 300,
  name: 'X-OAuth2-Authorization'
},
refreshToken: {
  property: 'refresh_token',
  maxAge: 1209600
},
```
Screen Shot 2020-10-06 at 10 09 10 am

packages/ripple-nuxt-tide/modules/preview/lib/helpers.js
tim-yao: Hey @alan-cole Not exactly like that. The previous preview is getting the section from route query already, you can just copy the implementation I think: https://github.com/dpc-sdp/ripple/blob/a7b310dc17a3b4d050191fdd2095ebff75e13e49/packages/ripple-nuxt-tide/modules/authenticated-content/lib/middleware.js#L41

examples/vic-gov-au/nuxt.config.js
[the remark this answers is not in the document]
tim-yao: Thanks @alan-cole I get what you mean. We can also move the client id setting to preview module just like this: https://github.com/dpc-sdp/ripple/blob/master/examples/vic-gov-au/nuxt.config.js#L137 , I feel that will make more sense.
But for now I would like just not add that config until we need to share this client id with other code. Just for reduce the maintenance work.

packages/ripple-nuxt-tide/modules/preview/pages/Login.vue
tim-yao: Some suggests avoid localStorge for security reason but if this is not sensible data, I am ok with it.
However, I think we may need to turn off it for auth module https://auth.nuxtjs.org/api/options.html#localstorage as it is using localStorage I can see.

packages/ripple-nuxt-tide/modules/preview/lib/helpers.js
alan-cole: Sorry, I was referring to these lines in terms of how site and section are used:
https://github.com/dpc-sdp/ripple/blob/master/packages/ripple-nuxt-tide/lib/core/tide.js#L419
https://github.com/dpc-sdp/ripple/blob/master/packages/ripple-nuxt-tide/lib/core/tide.js#L442

In my above code, I am using site section from query..

return getPageByPreviewLink(context.app.$tide, context.route.path, context.store.state.tide.siteId, **context.route.query.section**, headersConfig)

tim-yao: @alan-cole ahh, my bad, looked at wrong argument. It's correct then. One more thing, do you need to check if those context data are available? As chained object property can cause errors if the parent is undefined.

alan-cole: Yeah, that would be safest. I'll add some checks to prevent any breakages.

examples/vic-gov-au/nuxt.config.js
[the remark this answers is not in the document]
alan-cole: Reverted to previous implementation. Accesses env variable directly.

packages/ripple-nuxt-tide/modules/preview/lib/helpers.js
alan-cole: As per the above getPageByPreviewLink() discussion, this implementation should be what you're after.

packages/ripple-nuxt-tide/modules/preview/module.js
```diff
@@ -0,0 +1,67 @@
...
+          name: 'X-OAuth2-Authorization'
+        },
+        refreshToken: {
+          property: 'refresh_token',
+          maxAge: 60
+        },
+        responseType: 'code',
+        grantType: 'authorization_code',
+        clientId: options.oauth.clientId,
```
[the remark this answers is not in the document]
alan-cole: Reverted to using env variable directly.

packages/ripple-nuxt-tide/modules/preview/lib/helpers.js
alan-cole: My understanding is var `route.query` is always available, and if section isn't defined then it will return `undefined`.

Either

`const section = (context.route.query && context.route.query.section) ? context.route.query.section : null`

or

`const section = context.route.query.section`

will result in a falsy value and shouldn't make a difference to the execution and both should cause `pageData.section` to be null.

Similar to `context.store.state.tide.siteId`. I believe it's unlikely to be unset in practice.

Anyway, I've placed some checks just in case.

Maybe for a future improvement, as IE is the only browser left that doesn't support optional chaining, perhaps we could look at using:
https://babeljs.io/docs/en/next/babel-plugin-proposal-optional-chaining
to help clean up the codebase?

packages/ripple-nuxt-tide/modules/preview/pages/Login.vue
alan-cole: I've disabled localStorage use on the auth module.

tim-yao (approved):

packages/ripple-nuxt-tide/modules/preview/module.js
```diff
@@ -0,0 +1,71 @@
...
+
+  // Using `@nuxtjs/auth-next` instead of `@nuxtjs/auth` due to this bug:
+  // https://github.com/nuxt-community/auth-module/issues/761
+  // Switch to stable when v5 is released.
+  this.addModule(['@nuxtjs/auth-next', {
+    strategies: {
+      drupal: {
+        scheme: 'oauth2',
+        endpoints: {
+          authorization: `${options.baseUrl}oauth/authorize`,
```
tim-yao, line 28: Hey @alan-cole we need this to support baseUrl with and without slash in the end.

packages/ripple-nuxt-tide/modules/preview/lib/helpers.js
```diff
@@ -59,7 +59,8 @@ async function getPageByPreviewLink ($tide, path, site, section, headersConfig)
 async function tideShare (context, headersConfig) {
   const siteId = (context.store.state.tide && context.store.state.tide.siteId) ? context.store.state.tide.siteId : null
   const section = (context.route.query && context.route.query.section) ? context.route.query.section : null
-  return getPageByShareLink(context.app.$tide, context.route.path, siteId, section, headersConfig)
+  const path = context.route.path[0] === '/' ? context.route.path.substring(1) : context.route.path
```
tim-yao, line 62: Hi @alan-cole not sure why this... can you explain it please?

alan-cole: @tim-yao 

It's to avoid a double slash in the axios path. It's not necessary as the URL will still resolve, but it felt more correct to be passing well formed URLs than leaving it to axios to fix.
I'm happy to remove if you'd prefer the keep the code lean.

```js
const path = '/share_link/952b5923-7dce-4315-9b6a-2ead8e381be1/13358'

$tide.getByURL(path)

const apiPrefix = '/api/v1/'

function getByURL (url, headersConfig = {}) {
  return axios.$get(`${apiPrefix}${url}`, this._axiosConfig(headersConfig))
}

// axios request becomes:

// /api/v1//share_link/952b5923-7dce-4315-9b6a-2ead8e381be1/13358
//        ^
// This code removes this double //
```

tim-yao: Thanks, I am happy to keep your code. @alan-cole

PR #809, opened 2020-10-14
Update primary button active state

rubyllamadora:

### Changed

1.  Added primary button an active state as currently both primary and secondary share the same active state. This is so there's no need to create custom scss variable in dependent sites e.g. https://github.com/dpc-sdp/genderequalitycommission-vic-gov-au/pull/4/files#diff-65364271ce0ef6eb9c6d43d95c0ebdaedcf0c7d8954e741a83707fe88928c192R15


packages/components/Atoms/Global/scss/settings/_button.scss
```diff
@@ -19,8 +19,12 @@ $rpl-button-danger-hover-background-color: lighten(rpl-color('danger'), 5%) !def
 $rpl-button-danger-text-color: rpl_color('white') !default;
 $rpl-button-active-text-color: $rpl-button-text-color !default;
 $rpl-button-active-background-color: transparent !default;
-$rpl-button-active-background-image: rpl_gradient('primary_gradient_90') !default;
+$rpl-button-primary-active-background-image: rpl_gradient('primary_gradient_90') !default;
+$rpl-button-primary-active-background-color: transparent !default;
+$rpl-button-primary-active-background-image: $rpl-button-primary-active-background-image !default;
```
tim-yao: @rubyllamadora Duplicated `$rpl-button-primary-active-background-image` var? And it's assigned to itself? I think it's typo, isn't it?

tim-yao (changes requested):

packages/components/Atoms/Global/scss/settings/_button.scss
rubyllamadora: Updated :P

packages/components/Atoms/Global/scss/settings/_button.scss
```diff
@@ -19,8 +19,11 @@ $rpl-button-danger-hover-background-color: lighten(rpl-color('danger'), 5%) !def
 $rpl-button-danger-text-color: rpl_color('white') !default;
 $rpl-button-active-text-color: $rpl-button-text-color !default;
 $rpl-button-active-background-color: transparent !default;
-$rpl-button-active-background-image: rpl_gradient('primary_gradient_90') !default;
+$rpl-button-primary-active-background-color: transparent !default;
+$rpl-button-primary-active-background-image: rpl_gradient('primary_gradient_90') !default;
```
tim-yao, line 23: For backward compatibility, we may need to assign original value for this new one, so someone has set value for `$rpl-button-active-background-image` won't be impacted.
`$rpl-button-primary-active-background-image: $rpl-button-active-background-image !default;`.

BTW, do we need to add same stuff for secondary button?

rubyllamadora: Ugh good find. I'll update this.
Secondary button will stay as is?
`$rpl-button-secondary-active-background-image: $rpl-button-active-background-image !default;`

MdNadimHossain (approved):

PR #820, opened 2020-11-08
[SDPA-4787] iframe is not working in Accordion (Firefox only)

tim-yao:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-4787

### Changed

1.  Use custom css and js instead of v-show, so we can use visibility instead of `display:none` which triggers a Firefox bug to render iframe.


rubyllamadora (commented):
Hi @tim-yao, I have tested on smaller screen to check if it affects item with long content. The accordion item is now scrollable vertically, is this the desired behaviour?

packages/components/Organisms/Accordion/Accordion.vue
```diff
@@ -31,17 +28,14 @@
             <rpl-icon symbol="arrow_down_tertiary" color="primary" class="rpl-accordion__icon" :class="{'rpl-accordion__icon--expanded': accordionIsOpen(index)}"/>
           </button>
         </h2>
-        <transition
-          name="accordion"
-          @enter="start"
-          @after-enter="end"
-          @before-leave="start"
-          @after-leave="end"
```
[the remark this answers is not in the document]
tim-yao: Hey @rubyllamadora good point. It's because I am using height to do the accordion expend/collapse transition. Which has to be fixed number. When we do window resize, the fixed height need to be updated to avoid the scroll bar. I will have a look! Thanks!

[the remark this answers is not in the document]
tim-yao: Hey @rubyllamadora please review again. I have fixed the window resize height issue.

packages/components/Organisms/Accordion/Accordion.vue
```diff
@@ -88,23 +80,56 @@ export default {
...
+    collapseContent (index) {
+      const ref = this.accordionId(index)
+      const section = this.$refs[ref]
+      if (section[0]) {
+        // Set fixed height for transition
+        section[0].style.height = section[0].scrollHeight + 'px'
+        setTimeout(function () {
+          section[0].style.height = ''
+          section[0].style.visibility = ''
+        }, 1)
```
rubyllamadora, line 122: Interesting, this is like .001 seconds :P Do we still need to set timeout here?

rubyllamadora (approved):
@tim-yao, tested and looking good 👍

packages/components/Organisms/Accordion/Accordion.vue
tim-yao: This is an ugly work around to make the transition work. The reason is I need to have fixed height fist, then change it to 0. Without this 1 millisecond delay, the transition just not work for me. It's welcome if you have better way, but I can't find good solution based on the time for this.

rubyllamadora: Not really, just find it interesting how it works with that amount of millisecond waiting time.

PR #825, opened 2020-11-12
[SDPA-4714] card promo config

rubyllamadora:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-4714

### Changed

1.  Added Tide config to map card promo component
2. Moved common css styling for both nav and promo components into cards.scss to easily manage

### Screenshots

image

### DIFF
--- a/packages/components/Molecules/Card/CardPromo.vue
+++ b/packages/components/Molecules/Card/CardPromo.vue
@@ -14,7 +14,7 @@
               <span>{{status}}</span>
             </div>
           </slot>
-          <div class="rpl-card-promo__date" v-if="computedDate">{{ computedDate }}</div>
+          <div class="rpl-card-promo__date" v-if="formattedDate">{{ formattedDate }}</div>
         </slot>
       </div>
       <slot name="content">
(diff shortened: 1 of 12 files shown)


packages/components/Molecules/Card/CardPromo.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-  <rpl-link class="rpl-card-promo" :class="modifiers" :href="link.url">
```
tim-yao, line 2: We should keep `rpl-card-promo` class with modifiers if we follow `BEM`. See examples: http://getbem.com/naming/

packages/components/Molecules/Card/CardPromo.vue
```diff
@@ -38,15 +38,20 @@ export default {
...
+    image: Object,
+    dateStart: String,
+    dateEnd: String,
     tag: String,
     topic: String,
     status: String,
     title: String,
     summary: String,
-    link: Object
+    link: Object,
```
tim-yao, line 49: Does link a mandatory field for card? If it does, should we have a validation?

packages/components/Molecules/Card/CardPromo.vue
rubyllamadora: I've moved this to the modifiers() as it confuses me having two `class=""`. I'll put it back here so it's more clear that it's only for modifiers 👍

packages/components/Molecules/Card/CardPromo.vue
rubyllamadora: I probably need to review this props declaration e.g. provide default values if it's not mandatory and explicitly declare required fields.

tim-yao (approved):

rubyllamadora:
This has been QA'ed and sub-tickets created by QA for both Promo and Nav bugs. Merging to parent ticket.

PR #833, opened 2020-11-18
[SW-637] fix ripple icon issue with old svgs

rubyllamadora:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SW-637

### Changed

1.  Fixed svg icon not displaying on customIcon enabled sites


packages/components/Atoms/Icon/Icon.vue
```diff
@@ -1,6 +1,6 @@
 <!-- credits to https://github.com/Justineo/vue-awesome/blob/master/src/components/Icon.vue -->
 <template>
-  <svg v-if="icon" :class="iconClass" :style="iconStyle" :viewBox="box" aria-hidden="true" overflow="visible">
+  <svg v-if="icon || iconPropsSVG" :class="iconClass" :style="iconStyle" :viewBox="box" aria-hidden="true" overflow="visible">
```
tim-yao, line 3: @rubyllamadora Just feel the naming of `icon` and `iconPropsSVG` is not very helpful to understand what are they doing differently. Maybe can rename `iconPropsSVG` to something like `legacyIcon`(please use better name if you can find)?

If you can add a TODO comment in some place to tell us what to need to do to remove legacy code, that will be great. So we can do the job easier.

tim-yao (approved):

packages/components/Atoms/Icon/Icon.vue
rubyllamadora: I'm on board with legacyIcon, will update this and add more TODO comments. Thanks 👍

rubyllamadora: Updated and added TODO @tim-yao

tim-yao (approved):

PR #837, opened 2020-11-23
[SDPA-4647] update dependencies v2

rubyllamadora:

### Changed

1.  Updated patches/bug fixes and backward-compatible plugins using `yarn upgrade-interactive --latest` command.

### Screenshots
https://app-ripple-pr-837.lagoon.vicsdp.amazee.io/


Dockerfile.app
```diff
@@ -40,7 +40,6 @@ WORKDIR /app/examples/vic-gov-au/
 # force it to load the environment variable during build time. Otherwise it cannot read $LAGOON_GIT_BRANCH.
 RUN  . /home/.bashrc \
     && yarn run build --modern=client \
-    && chmod -R 755 ~/.config \
```
rubyllamadora, line 43: Lagoon failed building as this throws "not found" error. Removing this seems to completes the build fine. 
No build issue in CI.

packages/components/Molecules/Form/__tests__/RplSelect.test.js
```diff
@@ -29,8 +29,11 @@ describe('RplSelect', () => {
     ]
   }
 
+  const div = document.createElement('div')
+  document.body.appendChild(div)
+
   const baseConfig = {
-    attachToDocument: true,
+    attachTo: div,
```
rubyllamadora, line 36: Replaced to attachTo as attachToDocument is getting deprecated in the coming version.

packages/components/Molecules/Form/__tests__/RplSelect.test.js
```diff
@@ -60,68 +63,52 @@ describe('RplSelect', () => {
     expect(wrapper.find('.rpl-select__native').isVisible()).toBeTruthy()
   })
 
-  it('does not have native select on desktop', () => {
+  it('opens on click', async () => {
```
rubyllamadora, line 66: Changes below are just updating test with trigger events to use async await. Jest unit test fails on trigger events without having async await.

packages/ripple-nuxt-tide/modules/site/module.js
```diff
@@ -3,7 +3,7 @@ const path = require('path')
 module.exports = function () {
   const options = this.options.tide
   const { URL } = require('url')
-  const baseUrl = new URL(options.baseUrl)
+  const baseUrl = options.baseUrl ? new URL(options.baseUrl) : ''
```
rubyllamadora, line 6: I added extra validation as Lagoon build failed throwing FATAL error on `URL()` with invalid value in options.baseUrl.

tim-yao: I think we should throw an error and stop the build if `baseUrl` is not set, because without `baseUrl`, none of the Tide integration will work. We just need log a meaningful error message.  Maybe we should do validation for mandatory config option at the beginning of the build [(here)](https://github.com/dpc-sdp/ripple/blob/master/packages/ripple-nuxt-tide/lib/module.js#L8) How do you think?

tim-yao (approved):

packages/ripple-nuxt-tide/modules/site/module.js
rubyllamadora: Thanks for the suggestion @tim-yao. I'll look into this.

packages/ripple-nuxt-tide/lib/module.js
```diff
@@ -2,12 +2,17 @@ import defaults from './config/defaults'
...
 
 const path = require('path')
 
 const nuxtTide = function (moduleOptions) {
   const options = Object.assign(defaults, this.options.tide, moduleOptions)
 
+  if (!options || (options && !options.baseUrl)) {
+    logger.error('Base URL is not defined', { label: 'NuxtTide' })
+    throw new Error('Base URL is not defined')
+  }
```
rubyllamadora, line 15: @tim-yao not sure what are other mandatory data that I need to be checking. Is auth user/password required to have value?

tim-yao: @rubyllamadora There is another one `options.site`, but that one should be validated inside site module, not here.

packages/ripple-nuxt-tide/lib/module.js
```diff
@@ -2,12 +2,17 @@ import defaults from './config/defaults'
...
 import { RPL_HEADER } from './config/constants'
 import tideSearchApiMiddleware from '@dpc-sdp/ripple-tide-search-api'
+import logger from './core/logger'
 
 const path = require('path')
 
 const nuxtTide = function (moduleOptions) {
   const options = Object.assign(defaults, this.options.tide, moduleOptions)
 
+  if (!options || (options && !options.baseUrl)) {
```
tim-yao, line 12: @rubyllamadora I think you can use `if (!(options.baseUrl)) {`. Because `options` should be defined as object already in line 10.

rubyllamadora: ✔️ Will update.

PR #839, opened 2020-11-24
[SDPA-4764] open/close all accordion feature

rubyllamadora:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-4764

### Changed

1.  Added Open/Close all feature to accordion

### Screenshots
http://app-ripple-pr-839.lagoon.vicsdp.amazee.io/demo-landing-page

image


tim-yao (approved):

packages/components/Organisms/Accordion/Accordion.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <div class="rpl-accordion">
     <h2 class="rpl-accordion__title-top" :id="titleId" v-if="title">{{ title }}</h2>
+    <div class="rpl-accordion__collapse">
+      <a class="rpl-accordion__collapse-link" href="" @click.prevent="collapseExpandAll">{{ closeOpenLabel }}</a>
```
tim-yao, line 5: Hey @rubyllamadora just one minor thing, for better accessibility, this should be a button instead of link. You can find some explanation in `Buttons vs. Links` section in http://web-accessibility.carnegiemuseums.org/content/buttons/

tim-yao (approved):

packages/components/Organisms/Accordion/Accordion.vue
rubyllamadora: Good point. Was using link as I don't want to remove styling in button. Updating this to use `<button>`.

tim-yao: > Good point. Was using link as I don't want to remove styling in button. Updating this to use `<button>`.

@rubyllamadora yes, it was a design issue, it's better look like a button instead of link. If we can, may suggest back to designers.

PR #847, opened 2020-12-04
Feature/SDPA-4876 Added site section name hidden field for content rating form

MdNadimHossain:
**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-4876

### Changed

1.  Added a site section name prop for content rating form. Prop value will be passed from Tide.vue computed.
2. Added a hidden field for the content rating form, which will pass the prop value during form submission to the BE.


packages/ripple-nuxt-tide/lib/pages/Tide.vue
```diff
@@ -53,7 +53,7 @@
 
       <template slot="belowContent">
         <component class="rpl-site-constrain--on-all" v-if="campaignSecondary" :is="campaignSecondary.component" v-bind="campaignSecondary.data"></component>
-        <component :is="contentRating" class="rpl-site-constrain--on-all"></component>
+        <component :is="contentRating" :siteSectionName="siteSectionName" class="rpl-site-constrain--on-all"></component>
```
tim-yao, line 56: I guess below will work without need for `siteSectionName` method.
```suggestion
        <component :is="contentRating" :siteSectionName="page.appSection?.name" class="rpl-site-constrain--on-all"></component>
```
That's called `optional chaining operator` is a new way to dealing with undefined object properties.

packages/ripple-nuxt-tide/lib/pages/Tide.vue
```diff
@@ -224,6 +224,13 @@ export default {
         return () => import('@dpc-sdp/ripple-nuxt-tide/modules/webform/components/TideContentRating')
       }
       return false
+    },
+    siteSectionName () {
```
tim-yao, line 228: If above suggestion work, we don't need this I think.

packages/ripple-nuxt-tide/modules/webform/components/TideContentRating.vue
```diff
@@ -23,6 +23,9 @@ export default {
   components: {
     RplForm
   },
+  props: {
+    siteSectionName: String
```
tim-yao, line 27: Can we add a default value for it?
```
siteSectionName: {
  type: String,
  default: ''
}
```

https://vuejs.org/v2/guide/components-props.html#Prop-Validation

tim-yao (approved):

PR #849, opened 2020-12-07
[SDPSUP-2434] configurable timeout

tim-yao:

### Changed

1.  Added longer timeout for entity listing request
2. Make timeout configurable for `tide.get()`, so we can use different timeout setting if need
3. Expose timeout setting to nuxt tide module config


packages/ripple-nuxt-tide/README.md
```diff
@@ -33,6 +33,9 @@ You Can move some setting value into your env variables.
       password: 'yourPass' // Basic auth credential for Tide API.
     },
     site: 4, // Mandatory, should be your own site ID in Tide.
+    proxyTimeout: 60000, // Optional. Proxy request timeout. Default to 60s.
+    tideTimeout: 10000, // Optional. Tide API request timeout. Default to 10s.
+    tideListingTimeout: 30000, // Optional. Tide API listing request timeout. Default to 30s.
```
tim-yao, line 38: I made them configurable in `nuxt.config.js` now, instead of hardcode in code. But I didn't add them as env vars. I would like to add env vars when we really need. As it will need some extra work include CMDB and update all projects.

rubyllamadora (approved):

anthony-malkoun (approved):
This is easier to understand than the patch 🙂

PR #860, opened 2020-12-21
Release/1.19.2

rubyllamadora:

### Changed

1.


packages/create-ripple-app/sdp-releases.json
```diff
@@ -30,6 +30,6 @@
     "rippleVersion": "1.18.0"
   },
   "1.31.0": {
-    "rippleVersion": "1.19.1"
+    "rippleVersion": "1.19.2"
```
tim-yao, line 33: I think this commit is not got released, because it committed after Lerna release. Am I right? @rubyllamadora

rubyllamadora: Yes I pushed an empty release branch left my local change. That means I can't see the changes in npm?

PR #883, opened 2021-01-15
Ensure a default value is set so it doesn't lead to undefined error

rubyllamadora:
## Motivation and Context

## Changed

1. Getting component child error because of TypeError for undefined data - adding a default value solves the issue

Note: These are the only components I've found so far that affected `firerescue-vic-gov-au`. We should be consistent next time to add default values if we don't do defensive validation.

### Screenshots

image

## Types of changes

## Checklist

### DIFF
--- a/packages/components/Molecules/Form/__snapshots__/stories.storyshot
+++ b/packages/components/Molecules/Form/__snapshots__/stories.storyshot
@@ -1343,63 +1343,44 @@ exports[`RippleStoryshots Molecules/Form Default 1`] = `
                   Markup field
                 </h3>
                  
-                <figure
-                  class="rpl-document-link"
+                <article
+                  class="embedded-entity embedded-entity--media embedded-entity--media--document"
                 >
-                  <a
-                    aria-label=" Detailed Guide on the mandatory IR management criteria  File type: docx. Size: 75.22 KB"
-                    class="rpl-document-link__link"
-                    download=""
-                    href="/sites/default/files/2018-10/Detailed%20Guide%20on%20the%20mandatory%20IR%20management%20criteria.docx"
+                  <article
+                    class="media media--type-document media--view-mode-embedded"
                   >
-                    <svg
-                      aria-hidden="true"
-                      class="rpl-document-link__icon rpl-icon rpl-icon--docx rpl-icon--color_primary"
-                      overflow="visible"
-                      style="width: 30px; height: 30px;"
-                      viewBox="0 0 20 20"
-                    >
-                      <path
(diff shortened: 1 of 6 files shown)


packages/components/Molecules/Form/__snapshots__/stories.storyshot
```diff
@@ -1343,63 +1343,44 @@ exports[`RippleStoryshots Molecules/Form Default 1`] = `
...
+                <article
+                  class="embedded-entity embedded-entity--media embedded-entity--media--document"
                 >
-                  <a
-                    aria-label=" Detailed Guide on the mandatory IR management criteria  File type: docx. Size: 75.22 KB"
-                    class="rpl-document-link__link"
-                    download=""
-                    href="/sites/default/files/2018-10/Detailed%20Guide%20on%20the%20mandatory%20IR%20management%20criteria.docx"
+                  <article
+                    class="media media--type-document media--view-mode-embedded"
```
rubyllamadora, line 1350: An update to markup component seemed to have re-arranged all these snapshots changes.

tim-yao: @rubyllamadora it looks like the markup is plugin for document link is stop working in the PR. It should be convert to `rpl-document-link` in the story.

packages/components/Organisms/Grants/__snapshots__/stories.storyshot
```diff
@@ -411,39 +343,9 @@ exports[`RippleStoryshots Organisms/Grants Open - Countdown displayed 1`] = `
         </strong>
          with a 
         <a
-          class="rpl-link rpl-text-link"
-          data-print-url="https://vic.gov.au"
           href="https://vic.gov.au"
```
tim-yao, line 346: plugin for `rpl-link` is stop working as well.

tim-yao (changes requested):

packages/components/Molecules/Form/__snapshots__/stories.storyshot
rubyllamadora: So this is from the form stories here https://github.com/dpc-sdp/ripple/blob/develop/packages/components/Molecules/Form/stories.js#L356

If converted, it will become the previous `<figure class="rpl-document-link" ...`?
I thought it wasn't just the right snapshot before, I'll have a look. It's weird that having a recommended default value to props will cause this.

rubyllamadora:
@tim-yao I'm closing this as I figured out the root cause that is inside cheerio module https://github.com/dpc-sdp/firerescue-vic-gov-au/pull/17/files#diff-e9ac4a92db8f7d70e5a3b775113d9f3f60ebab76b7e9cea75748b4b2e938b8bfR10

PR #888, opened 2021-01-22
[sdpsup-2507] preview error

tim-yao:
## Motivation and Context

**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDSUP-2507

## Changed

1. Handling the several nuxt auth errors better

## How Has This Been Tested?

none, e2e test will be added in nightwatch

## Types of changes

## Checklist

### DIFF
--- a/packages/ripple-nuxt-tide/lib/templates/plugin.js
+++ b/packages/ripple-nuxt-tide/lib/templates/plugin.js
@@ -93,9 +93,18 @@ export default ({ app, req, store , route }, inject) => {
         },
         async setSiteData ({ commit }, { requestId = null } = {}) {
           const headersConfig = { requestId }
-          const siteData = await app.$tide.getSiteData(headersConfig)
+          let siteData = await app.$tide.getSiteData(headersConfig)
           if (siteData instanceof Error) {
-            throw siteData
+            // If got a Nuxt auth session expire error, try again.
+            // As Nuxt auth should reset auth on error, retry should pass.
+            if (siteData.name === 'ExpiredAuthSessionError') {
+              siteData = await app.$tide.getSiteData(headersConfig)
+              if (siteData instanceof Error) {
+                throw siteData
+              }
+            } else {
+              throw siteData
+            }
           }
           siteData.lastFetched = Date.now()
           commit('setSiteData', siteData)
--- a/packages/ripple-nuxt-tide/modules/preview/templates/plugin.js
+++ b/packages/ripple-nuxt-tide/modules/preview/templates/plugin.js
@@ -1,9 +1,18 @@
-export default function ({ $axios, app, redirect }) {
+import { logger } from '@dpc-sdp/ripple-nuxt-tide/lib/core'
+
+export default function ({ $auth, $axios }) {
+
+  $auth.onError((error, name, endpoint) => {
+    if (process.server) {
+      logger.debug('Auth error: "%O"', name, { error, label: 'Tide preview' })
+    }
+  })
 
   $axios.onError(error => {
-    // Redirect to login page again if both token and refresh token have expired.
-    if (error.name?.includes('ExpiredAuthSessionError')) {
-      redirect('/oauth/login')
+    // Sometime BE will return login page instead of json error response if token is invalid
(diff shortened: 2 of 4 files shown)


packages/ripple-nuxt-tide/modules/preview/templates/plugin.js
```diff
@@ -1,9 +1,18 @@
...
+  $auth.onError((error, name, endpoint) => {
+    if (process.server) {
+      logger.debug('Auth error: "%O"', name, { error, label: 'Tide preview' })
+    }
+  })
 
   $axios.onError(error => {
-    // Redirect to login page again if both token and refresh token have expired.
-    if (error.name?.includes('ExpiredAuthSessionError')) {
-      redirect('/oauth/login')
```
MdNadimHossain, line 6: @tim-yao general question? If we are not redirecting, then what will happen after the expired auth session?

packages/ripple-nuxt-tide/modules/preview/templates/plugin.js
```diff
@@ -1,9 +1,18 @@
...
+    // Sometime BE will return login page instead of json error response if token is invalid
+    if (error.response?.status === 401 && error.response?.data?.error === undefined) {
+      logger.debug('Aixos error with no proper error, e.g HTML login page', { error, label: 'Tide preview' })
```
rubyllamadora, line 14: Does it have to be always a 401 and undefined error?
Also *Axios (:  and should this be Axios returned unauthorised error or something since it is 401?

packages/ripple-nuxt-tide/lib/templates/plugin.js
```diff
@@ -93,9 +93,18 @@ export default ({ app, req, store , route }, inject) => {
...
         async setSiteData ({ commit }, { requestId = null } = {}) {
           const headersConfig = { requestId }
-          const siteData = await app.$tide.getSiteData(headersConfig)
+          let siteData = await app.$tide.getSiteData(headersConfig)
           if (siteData instanceof Error) {
-            throw siteData
+            // If got a Nuxt auth session expire error, try again.
+            // As Nuxt auth should reset auth on error, retry should pass.
+            if (siteData.name === 'ExpiredAuthSessionError') {
+              siteData = await app.$tide.getSiteData(headersConfig)
```
rubyllamadora, line 101: The extending of timeout didn't work? I feel like the main issue lies in the API somewhere and this is just to have a workaround.

rubyllamadora (approved):
Hi @tim-yao, looks fine to me. I just find the re-request interesting as it already has enough timeout time.

packages/ripple-nuxt-tide/modules/preview/templates/plugin.js
tim-yao: It should handled by auth module [resetonerror](https://auth.nuxtjs.org/api/options/#resetonerror) I have it enabled https://github.com/dpc-sdp/ripple/blob/develop/packages/ripple-nuxt-tide/modules/preview/module.js#L49

But that does not seem work for some case, that's why I have it in https://github.com/dpc-sdp/ripple/pull/888/files#diff-ae894b114c4ab959cac6d35451789a5ca617aa59996151f7e0bffa6392172336R100 to retry. So far looks it handle all the errors in our tests.

MdNadimHossain (approved):

packages/ripple-nuxt-tide/modules/preview/templates/plugin.js
tim-yao: Tide should give error response but in some API request, Tide gives a redirect to login page with 401 error, instead of oauth error. I saw some discuss for that before, I think that is a thing hard to be changed in Tide for now. So this is a patch for that issue.

packages/ripple-nuxt-tide/lib/templates/plugin.js
tim-yao: @rubyllamadora this is for a handle the auth token expiration error not about timeout, see https://github.com/dpc-sdp/ripple/pull/888#discussion_r562401166

PR #889, opened 2021-01-25
Install curl for all environments.

nicksantamaria:


anthony-malkoun:
I don't see any harm in installing curl in all environments and makes for a cleaner Dockerfile.

nicksantamaria:
Just checked, this is essentially how consumer ripple sites are set up anyway https://github.com/dpc-sdp/vic-gov-au/blob/develop/Dockerfile#L36

tim-yao (approved):

PR #894, opened 2021-02-02
[SDPA-4969] Added fix for webform number field states.

MdNadimHossain:
## Motivation and Context

**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-4969

## Changed

1. Added field type check in conditional logic, if the input type is a number then calculate the states based on the number field.

## Types of changes

## Checklist


packages/ripple-nuxt-tide/modules/webform/conditional-logic.js
```diff
@@ -147,19 +147,38 @@ function performTriggerCheck (rule) {
       if (typeof rule.triggerValue === 'string') {
```
tim-yao, line 147: I guess this logic will filter out all number values. So new logic https://github.com/dpc-sdp/ripple/pull/894/files#diff-5667abf8c7db140ddc6fc57e232e1a5f2268574feb08a3db960b2b3d87baef40R150 won't be used?

packages/ripple-nuxt-tide/modules/webform/conditional-logic.js
```diff
@@ -147,19 +147,38 @@ function performTriggerCheck (rule) {
...
         // value
         result = (rule.modelValue === rule.triggerValue)
+        if (rule.modelValue && fieldType != null && fieldType === 'number') {
```
tim-yao, line 150: Will use `typeof` to check if number work instead of pass `fieldType` through several level of functions?

tim-yao (commented):
Hi @MdNadimHossain have some question in the comment. Please have a look. thanks!

packages/ripple-nuxt-tide/modules/webform/conditional-logic.js
MdNadimHossain: @tim-yao This logic filters out the value that is set in the BE, against which the entered Value is matched. The value that gets sends from the BE is always a string even if you set a number value. This is why I am converting the trigger value to a number to match with the entered number value.

packages/ripple-nuxt-tide/modules/webform/conditional-logic.js
MdNadimHossain: right, I am updating this.

packages/ripple-nuxt-tide/modules/webform/conditional-logic.js
```diff
@@ -126,14 +126,14 @@ function performTest (test) {
...
-function performTriggerCheck (rule) {
+function performTriggerCheck (rule, fieldType) {
   let result = true
   switch (rule.triggerName) {
     case 'empty':
       result = (rule.modelValue == null || rule.modelValue === '')
       break
     case 'filled':
-      result = (rule.modelValue != null && rule.modelValue.length > 0)
+      result = (rule.modelValue && rule.modelValue != null) ? (rule.modelValue && rule.modelValue != null) : false
```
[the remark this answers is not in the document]
MdNadimHossain: ahh good catch, could have just used the same logic from the `case empty` and reverse it. Thanks

tim-yao (approved):

PR #901, opened 2021-02-11
Updated log client to read environment variables with sane defaults.

nicksantamaria:
This fixes an issue with application logs not being aggregated on the new Azure cluster.

### DIFF
--- a/packages/ripple-nuxt-tide/lib/core/logger.js
+++ b/packages/ripple-nuxt-tide/lib/core/logger.js
@@ -95,9 +95,7 @@ const logstash = new LogstashTransport({
 let logger = createLogger({
   level: logLevel,
   format: format.combine(
-    format.timestamp({
-      format: 'YYYY-MM-DD HH:mm:ss'
-    }),
+    format.timestamp(),
     format.splat()
   ),
   defaultMeta: { service: 'ripple-tide' },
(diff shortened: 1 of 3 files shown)


packages/ripple-nuxt-tide/lib/core/logger.js
```diff
@@ -53,19 +53,17 @@ const errorPrint = format(info => {
...
   const gitBranch = process.env.LAGOON_GIT_SAFE_BRANCH || LAGOON_LOGS_DEFAULT_SAFE_BRANCH
-  const type = [openshiftProject, gitBranch]
+  const type = [k8sNamespace, gitBranch]
   info.type = type.join('-')
   return info
 })
 
 // Set up level based on env settings.
-let logLevel = 'info'
+let logLevel = process.env.LOG_LEVEL || 'warn'
```
tim-yao, line 64: 👍

tim-yao (approved):
Thanks @nicksantamaria , looks good to me.

nicksantamaria:
@tim-yao I was looking at the possibility of refactoring how the log transports were added, so the console was only used for local dev + clientside. But never got it working. 


Lets go with this PR as-is. Merge if you're happy with it.

tim-yao:
> @tim-yao I was looking at the possibility of refactoring how the log transports were added, so the console was only used for local dev + clientside. But never got it working.
> 
> Lets go with this PR as-is. Merge if you're happy with it.



> @tim-yao I was looking at the possibility of refactoring how the log transports were added, so the console was only used for local dev + clientside. But never got it working.
> 
> Lets go with this PR as-is. Merge if you're happy with it.

@nicksantamaria Thanks. I will have a quick look to see if I can disable console log for lagoon in this PR. If not I will merge it.

PR #964, opened 2021-05-20
[SDPA-4924] Fix Link text doesn't wrap in accordions

mayurngondhkar:
## Motivation and Context

**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-4924

## Changed

1. Link text doesn't wrap

## Types of changes

## Checklist

### DIFF
--- a/packages/components/Atoms/Icon/TextIcon.vue
+++ b/packages/components/Atoms/Icon/TextIcon.vue
@@ -7,6 +7,10 @@
   <span v-else-if="text && symbol && placement === 'after'">
     <span v-if="textWordCount > 1">{{ textWithoutLastWord }}</span>
     <span v-if="textWordCount > 1" class="rpl-text-icon__group"> {{ textLastWord }}<rpl-icon v-bind="iconProps" /></span>
+    <span v-if="textWordCount === 1 && textIsAbsoluteUrl">
+      <div class="rpl-text-icon__absolute-url">{{ text }}</div>
+      <div class="rpl-text-icon__icon"><rpl-icon v-bind="iconProps" /></div>
+    </span>
     <span v-else class="rpl-text-icon__group">{{ text }}<rpl-icon v-bind="iconProps" /></span>
   </span>
   <span v-else-if="text">{{ text }}</span>
@@ -14,6 +18,7 @@
 
 <script>
 import RplIcon from './Icon.vue'
+import { isAbsoluteUrl } from "@dpc-sdp/ripple-global/utils/helpers"
 
 /**
  * Text Icon allows user to display text with an icon that can be placed before or after.
@@ -63,6 +68,9 @@ export default {
         size: this.size,
         class: `rpl-text-icon--${this.placement}`
       }
+    },
+    textIsAbsoluteUrl: function () {
+      return isAbsoluteUrl(this.text)
     }
   }
 }
@@ -85,6 +93,19 @@ export default {
 
     &--after {
       margin: $rpl-text-link-after-margin;
+    }
+
+    &__absolute-url {
+      display: block;
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
+    }
+
+    &__icon {
+      display: block;
+      float: right;
+      margin: -22px -13px 0px 0px;
     }
   }
 </style>
--- a/packages/components/Atoms/Icon/TextIcon.vue
+++ b/packages/components/Atoms/Icon/TextIcon.vue
@@ -5,9 +5,11 @@
     <span v-else class="rpl-text-icon__group"><rpl-icon v-bind="iconProps" />{{ text }}</span>
   </span>
   <span v-else-if="text && symbol && placement === 'after'">
-    <span v-if="textWordCount > 1">{{ textWithoutLastWord }}</span>
-    <span v-if="textWordCount > 1" class="rpl-text-icon__group"> {{ textLastWord }}<rpl-icon v-bind="iconProps" /></span>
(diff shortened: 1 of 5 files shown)


mayurngondhkar:
Hi @rubyllamadora,

Can you please review this PR.

https://app.pr-964.ripple.sdp2.sdp.vic.gov.au/
https://app.pr-964.ripple.sdp2.sdp.vic.gov.au/clone-demo-landing-page

Image:
image

image


https://digital-engagement.atlassian.net/browse/SDPA-4924



Thanks

packages/components/Atoms/Icon/TextIcon.vue
```diff
@@ -5,15 +5,22 @@
...
   </span>
   <span v-else-if="text && symbol && placement === 'after'">
-    <span v-if="textWordCount > 1">{{ textWithoutLastWord }}</span>
-    <span v-if="textWordCount > 1" class="rpl-text-icon__group"> {{ textLastWord }}<rpl-icon v-bind="iconProps" /></span>
+    <span v-if="textWordCount > 1">
+      <span>{{ textWithoutLastWord }}</span>
+      <span class="rpl-text-icon__group"> {{ textLastWord }}<rpl-icon v-bind="iconProps" /></span>
+    </span>
+    <span v-else-if="textWordCount === 1 && textIsAbsoluteUrl" class="rpl-text-icon__url-icon">
+      <div class="rpl-text-icon__absolute-url">{{ text }}</div>
```
rubyllamadora, line 13: Hmm this is tricky as text props can accept a group of words or just URL or both. Can you try this css setting 
```
overflow-wrap: break-word;
word-wrap: break-word;
```
It should break long words and respect the parent's width.
I found sample code in here https://codepen.io/team/css-tricks/pen/RWaNxr

rubyllamadora: Also can we update unit test with this test case please `text-icon.test.js`

packages/components/Atoms/Icon/TextIcon.vue
```diff
@@ -86,5 +102,21 @@ export default {
     &--after {
       margin: $rpl-text-link-after-margin;
     }
+
+    &__url-icon {
+      display: flex;
+    }
+
+    &__absolute-url {
+      display: block;
```
rubyllamadora, line 111: This will bring down the whole text to the next line. Try inline-block?

rubyllamadora (changes requested):
Hey @mayurngondhkar, good effort on trying to fix the issue however I'm still not convinced to add custom absolute_url validation to the component. I feel like the component should stand to its Text+Icon. And the main issue here is when the TextIcon is wrapped in an Accordion/Markup - the URL just doesn't break or respect the max-width set in markup__inner.

Can you try removing 
```
&__group {
  white-space: nowrap;
}
```
in https://github.com/dpc-sdp/ripple/pull/964/files#diff-5744049b83242975493c5079979620b8d1994802977a578977a9e9a1956c2aefR95

Hopefully it works so there's no need to add custom validation (:

packages/components/Atoms/Icon/TextIcon.vue
dylankelly: @rubyllamadora @mayurngondhkar Not required, but I reckon this is a good candidate for a mixin - See Chris Coyiers https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/#for-the-scss-inclined

`@mixin rpl_word_wrap`

packages/components/Atoms/Icon/TextIcon.vue
```diff
@@ -63,6 +76,9 @@ export default {
         size: this.size,
         class: `rpl-text-icon--${this.placement}`
       }
+    },
+    textIsAbsoluteUrl: function () {
```
dylankelly, line 80: if `isAbsoluteUrl` is a function, no need to wrap in another function. 
```
<template>
<span v-if="isAbsoluteUrl(text)">something</span> 
</template>
<script>
methods: {
  isAbsoluteUrl,
}
</script>
```

dylankelly (changes requested):
> Hey @mayurngondhkar, good effort on trying to fix the issue however I'm still not convinced to add custom absolute_url validation to the component. I feel like the component should stand to its Text+Icon. And the main issue here is when the TextIcon is wrapped in an Accordion/Markup - the URL just doesn't break or respect the max-width set in markup__inner.
> 
> Can you try removing
> 
> ```
> &__group {
>   white-space: nowrap;
> }
> ```
> 
> in https://github.com/dpc-sdp/ripple/blob/develop/packages/components/Atoms/Icon/TextIcon.vue#L78
> 
> Hopefully it works so there's no need to add custom validation (:

I agree with @rubyllamadora - not convinced this is needed.

packages/components/Atoms/Icon/TextIcon.vue
mayurngondhkar: I spoke to Jess and she was ok with the text wrapping to the next line.

Thanks @dylankelly and @rubyllamadora

mayurngondhkar:
@rubyllamadora please have a look at the PR when you get a chance. Thanks

packages/components/Atoms/Icon/TextIcon.vue
dylankelly: @mayurngondhkar have another look at my comment - im talking about how you are adding this function into the vue component

dylankelly: ok i see you have removed this 👍

rubyllamadora (approved):

PR #973, opened 2021-06-21
Extended shareThis component to support more networks

anfilimon:
## Motivation and Context

While vue-social-sharing supports multiple networks, the current ShareThis component is specifically looking for and rendering the 3 default values.
This change makes the component a bit more dynamic, to support other networks as well.

## Changed

1. Replaced the if statements looking for the 3 default networks with a dynamic for loop
2. Added the label to the computed properties since it is used inside the social-sharing tag
3. Added an extra field for email (please see notes below for more info)
4. Changed the network tag from button to anchor, for consistency with email
5. Added tests
6. Updated the stories
7. Added whatsapp icon in ripple-icon (please see notes below for more info)

**Notes:**

- At the moment, we only share the URL with no extra title or description. Most networks can use the page's meta data to extract any missing information. However, emails require a specific subject and body to work as expected. Since language is also often different in emails (compared to social media), custom information might be better. This is why I have split the email from the dynamic for loop.
- For the new whatsapp icon - I tried updating shareThis in my node_modules and simply adding the icon in my project to test, but didn't seem to work. However, adding it to ripple-icon works fine. Let me know if you can think of a better way, so projects can use any network supported by vue-social-sharing without updating ripple-icon.

### Screenshots
**Default:**
Screen Shot 2021-06-21 at 11 13 51 am

**With Email:**
Screen Shot 2021-06-21 at 11 14 00 am

**With Custom network:**
Screen Shot 2021-06-21 at 11 14 08 am

## How Has This Been Tested?

- Added unit tests
- Visual/manual testing in storybook

## Types of changes

## Checklist

- [x] I've added relevant changes to the documentation.
- [x] I have added tests to cover my changes (if not applicable, please state why)


packages/components/Atoms/Icon/icons/index.js
```diff
@@ -86,6 +86,7 @@ import './upload'
 import './user'
 import './view'
 import './webinar'
+import './whatsapp'
```
dylankelly, line 89: Adding Whatsapp icon here is not so bad, but note projects can register their own custom icons by placing them in `/assets/ripple-icon` folder - See https://github.com/dpc-sdp/ripple/blob/f177592f80561824b30350cebf48b9c74c0dd7bc/examples/basic-examples/assets/ripple-icon/custom_icon_search.js  And webpack loader config https://github.com/dpc-sdp/ripple/blob/develop/packages/ripple-nuxt-ui/lib/module.js

packages/components/Molecules/ShareThis/__test__/ShareThis.test.js
```diff
@@ -0,0 +1,72 @@
+import { shallowMount } from '@vue/test-utils'
```
dylankelly, line 1: Thanks for adding tests!

dylankelly (approved):
Thank you for contributing this @anfilimon ! I'm happy for this to be merged pending successful QA. Regarding the print button, I think this should remain a seperate component. Happy for you to contribute this as another component if you wish. I would like to see it as a configurable option in Drupal sidebar options to enable the print component.

packages/components/Atoms/Icon/icons/index.js
anfilimon: Hey @dylankelly! thanks for the links - I did try to add it to /assets/ripple-icon; it worked fine if referenced inside my project, under /components for example, but didn't render when referenced from inside a ripple dependency. I was probably missing something in webpack, so will have a look at that.

dylankelly:
@anfilimon I've moved this to #973 as our lagoon infrastructure doesn't support building from forks - closing this one.

PR #981, opened 2021-06-29
[SDPA-5312] Print this component

dylankelly:
## Motivation and Context
Same functionality and design as Print this page option in https://www.ripple.sdp.vic.gov.au/?path=/story/organisms-publication--publication-download-print but is rendered based on a checkbox in sidebar, same as social share.

**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-5312

## Changed

1. Added new package called print-this
2. Updated docs

### Screenshots
demo-landing-page-with-print

**Note**: our requirements are to have print above social sharing, so we had to update the sidebar order.

## How Has This Been Tested?
Manual UI testing in storybook

## Types of changes

## Checklist

I'm not sure whether to tick the template changes - it's not a breaking change but I imagine it would need to be added to the sidebar middleware

- [x] I've added relevant changes to the documentation.


dylankelly:
@anfilimon This has the same problem of not working properly in safari - See webkit bug https://bugs.webkit.org/show_bug.cgi?id=195769

The solution there is to use `document.execCommand('print', false, null);` See - https://stackoverflow.com/a/50473614
Do you think you could try and add this?

anfilimon:
> @anfilimon This has the same problem of not working properly in safari - See webkit bug https://bugs.webkit.org/show_bug.cgi?id=195769
> 
> The solution there is to use `document.execCommand('print', false, null);` See - https://stackoverflow.com/a/50473614
> Do you think you could try and add this?

Hey @dylankelly just FYI that document.execCommand('print') doesn't work in firefox and needs a support check and fallback. The update seems to work for all major browsers. I wasn't sure if you wanted to extend the scope of this ticket to the publication, so left the fix out there.

dylankelly:
> > @anfilimon This has the same problem of not working properly in safari - See webkit bug https://bugs.webkit.org/show_bug.cgi?id=195769
> > The solution there is to use `document.execCommand('print', false, null);` See - https://stackoverflow.com/a/50473614
> > Do you think you could try and add this?
> 
> Hey @dylankelly just FYI that document.execCommand('print') doesn't work in firefox and needs a support check and fallback. The update seems to work for all major browsers. I wasn't sure if you wanted to extend the scope of this ticket to the publication, so left the fix out there.

Wonderful this is great. I'll look at incorporating this button into the publication component somehow later. Might need a design change.

PR #1012, opened 2021-08-02
[SDPA-5406] Removed the summary field reference for news attached to the body field.

MdNadimHossain:
## Motivation and Context

**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-5406

## Changed

1. Removed the summary field mapping from the filter, so it does not match the summary field which is attached to the body field. The new changes in the BE has removed that field for news content type and from now all the summary field should reference to the single one which is field_landing_page_summary.

### Screenshots
Screen Shot 2021-08-02 at 1 20 22 pm

## Types of changes

## Checklist


mayurngondhkar (approved):
LGTM

PR #1032, opened 2021-09-21
SDPA-5631: External icon link accesible.

mudrasamey:
## Motivation and Context

**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-5631

## Changed

1. Added alt text, text and changed aria-hidden for the external link icon so that its visible to the screen reader

### Screenshots

Screenshot 2021-09-20 at 11 49 24 AM


packages/components/Atoms/Icon/Icon.vue
```diff
@@ -87,6 +88,22 @@ export default {
...
+      }
+    },
+    ariaHidden () {
+      if (!this.icon && !this.legacyIcon) return 'TRUE'
+      if (this.symbol === 'external_link') {
+        return 'FALSE'
+      } else {
+        return 'TRUE'
+      }
+    },
```
elNico, line 106: Hey @mudrasamey, had another look, could you maybe combine those 2 methods and set those 2 properties at the same time? The checks are identical in both, so it's mostly duplication. 

Can you also have a look through the codebase for other examples where `aria-hidden` is used and see if it's all caps and if not use whatever casing is already used. Thanks!

elNico (changes requested):
Pls see comment.

packages/components/Atoms/Icon/Icon.vue
mudrasamey: @elNico I have removed the duplicate method and replaced the aria-hidden to lowercase to match rest of the implementations

dylankelly (approved):

packages/components/Atoms/Icon/Icon.vue
dylankelly: @elNico I see this has been implemented - can you look at this again?

dylankelly (approved):

elNico (approved):

PR #1039, opened 2021-09-26
SDPA-5759: Adding aria-live attribute to alerts.

mudrasamey:
## Motivation and Context
Alerts in SDP are dynamically loaded and hence we would to prioritise the content while screen reading as soon as the content appears on the page. The idea here is to use aria-live attribute as mentioned https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-5759

## Changed

1. Adding aria-live attribute to alerts.


dylankelly:
@mudrasamey Thanks for the PR - Please fill out the PR template with more detail about the motivation for this change - the linked ticket has no details either.

mudrasamey:
> @mudrasamey Thanks for the PR - Please fill out the PR template with more detail about the motivation for this change - the linked ticket has no details either.

mudrasamey:
@dylankelly ive added more details as mentioned

dylankelly (approved):
Thanks for providing context @mudrasamey - I'm happy to pass this, would appreciate if you can provide detailed instructions for testing on the Jira ticket.

PR #1040, opened 2021-09-27
SDPA-5649: Adding aira labeled by attribute to card icon.

mudrasamey:
## Motivation and Context

**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-5649

## Changed

1. Adding aria-labelledby attribute to the card icon to improve accessibility


packages/components/Molecules/Card/CardEmergencyContact.vue
```diff
@@ -2,7 +2,7 @@
   <div class="rpl-card-emergency">
     <div class="rpl-card-emergency__content">
       <div class="rpl-card-emergency__subtitle" v-if="subtitle"><span>{{ subtitle }}</span></div>
-      <h2 class="rpl-card-emergency__title" v-if="title">{{ title }}</h2>
+      <h2 class="rpl-card-emergency__title" v-if="title" :for="id" :id="id" >{{ title }}</h2>
```
elNico, line 5: @mudrasamey what is the ":for" for?

packages/components/Molecules/Card/CardEmergencyContact.vue
```diff
@@ -40,6 +40,14 @@ export default {
     RplIcon,
     RplTextIcon
   },
+  data () {
+    return {
+      id: null
+    }
+  },
+  mounted () {
+    this.id = 'card_title_' + Math.random().toString(36).substr(2, 9)
```
elNico, line 49: @mudrasamey better use dashes instead of underscores, so "card-title-"

packages/components/Molecules/Card/CardEmergencyContact.vue
```diff
@@ -13,7 +13,7 @@
         :innerWrap="false"
         v-if="link"
       >
-        <rpl-icon :symbol="iconSymbol" color="primary" size="1.05"/>
+        <rpl-icon :symbol="iconSymbol" color="primary" size="1.05" :aria-labelledby="id"/>
```
elNico, line 16: In this case we don't want to put this on the icon I think (the external one was a special case)...here it should probably go on the anchor tag itself, meaning we also need to modify that and pass it on from here.

elNico (changes requested):
A few points to address

packages/components/Molecules/Card/CardEmergencyContact.vue
```diff
@@ -12,6 +12,7 @@
         :href="filterLink"
         :innerWrap="false"
         v-if="link"
+        :aria-labelledby="id"
```
elNico, line 15: @mudrasamey rpl-link hasn't got this yet as a property, though, so it needs to be added there as well and passed on:
https://github.com/dpc-sdp/ripple/blob/develop/packages/components/Atoms/Link/Link.vue

elNico (changes requested):
The Link component needs to be modified as well.

elNico (approved):

packages/components/Molecules/Card/CardEmergencyContact.vue
```diff
@@ -12,6 +12,7 @@
         :href="filterLink"
         :innerWrap="false"
         v-if="link"
+        :ariaLabelledBy="id"
```
dylankelly, line 15: See comment on link component: you can just add `aria-labelledby` here without having to pass it as a prop and it will be added to the root element of the link component (the a element in this case)

dylankelly (changes requested):
Thanks for this - have requested a few changes.

dylankelly (approved):
Looks great @elNico , appreciate you addressing those comments. Happy to approve this. Can I get you to move the Jira ticket through?

PR #1046, opened 2021-10-04
SDPA-5652: Adding aria-labelledby to alert.

mudrasamey:
## Motivation and Context

While screen reading the alert link doesn't have a suitable context. 
Proposed solution is to use aria-labelledby to reference alert text.

**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-5652

## Changed

1. Added aria-labelledby attribute to the  alert link referencing the alert text.

### Screenshots
Screenshot-2021-10-04-at-12-34-06-PM


dylankelly (approved):
Thanks @mudrasamey @elNico looks good.

PR #1122, opened 2022-01-18
[RPL2-129] Map Grant content type to new API

waitingallday:
## Motivation and Context

**JIRA issue:** https://digital-engagement.atlassian.net/browse/RPL2-129

## Changed

1. Created `ripple-tide-grant`
2. Mapped fields from new API to Vue SFC

## Types of changes

## Checklist


packages/ripple-tide-grant/index.js
```diff
@@ -0,0 +1,89 @@
...
+      enddate: getField(src, 'field_node_dates.end_value', ''),
+      description: getField(src, 'field_description.processed', ''),
+      link: getLinkFromField(src, 'field_call_to_action')
+    }),
+    timeline: (src) => ({
+      title: getField(src, 'field_node_timeline.field_paragraph_title'),
+      list: getField(src, 'field_node_timeline.field_timeline').map(timeline => ({
+        title: getField(timeline, 'field_paragraph_title'),
+        subtitle: getField(timeline, 'field_paragraph_cta_text'),
+        url: timeline.field_paragraph_link ? timeline.field_paragraph_link.origin_url || timeline.field_paragraph_link.uri : null,
```
waitingallday, line 59: Might be worth adding the extra fields from this example to `getLinkFromField` - I wasn't sure if this was specific to the timeline data structure

packages/ripple-tide-grant/index.js
```diff
@@ -0,0 +1,89 @@
...
+        image: timeline.field_paragraph_media && timeline.field_paragraph_media.field_media_image ? timeline.field_paragraph_media.field_media_image.url || timeline.field_paragraph_media.field_media_image.uri : null,
```
waitingallday, line 60: Same with this one - `getImageFromField` isn't set up to deal with this example

packages/ripple-tide-grant/index.vue
```diff
@@ -0,0 +1,148 @@
+<template>
```
waitingallday, line 1: I used `ripple-tide-landing-page` as a template for this, not sure how much of this markup is relevant?

packages/ripple-tide-grant/index.js
```diff
@@ -0,0 +1,89 @@
+import mime from 'mime-types'
+import { getField, getLinkFromField } from '@dpc-sdp/ripple-tide-api/src/services/utils'
+import components from './component-loader'
+
+// Utils
+const formatAudiences = (audiences) => {
```
waitingallday, line 6: This one seems very specific to parsing a single word out of a giant object, I don't think it needs to go out to utils

packages/ripple-tide-grant/index.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+  }
+  if (audiences && audiences.length > 0) {
+    const audienceStr = [...new Set(audiences)].map(a => formatAudience(a)).join(', ')
+    return `${audienceStr.charAt(0).toUpperCase() + audienceStr.slice(1)}`
+  }
+  return ''
+}
+
+const formattedSize = (fileSize) => {
```
waitingallday, line 27: This is probably a good candidate to go into utils

packages/ripple-tide-grant/component-loader.js
```diff
@@ -0,0 +1,3 @@
+export default {
```
dylankelly, line 1: As there are no dynamic components for this content type, we can remove the component loader and component-mapping files

packages/ripple-tide-grant/index.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+  return ''
+}
+
+export default {
+  pageComponent: () => import(/* webpackMode: "eager" */ '@dpc-sdp/ripple-tide-grant/index.vue'),
+  components,
+  mapping: {
+    title: 'title',
+    overview: (src) => ({
```
dylankelly, line 45: I've added support for object syntax now so you should be able to refactor this to just use the string path in these nested fields

packages/ripple-tide-grant/definition.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - bodyComponents
+  - headerComponents
+allOf:
+  - $ref: '#/components/schemas/TidePage'
+properties:
+  type:
+    type: string
+    enum:
+      - grant
+  title:
```
dylankelly, line 16: Will need to add the fields you've added to this schema too. See https://json-schema.org/understanding-json-schema/index.html for syntax. I'd like to explore how we might be able to move this to an annotation based setup instead of having to do this eg: https://github.com/john-doherty/jsdoc-to-json-schema#input-singleton

packages/ripple-tide-grant/index.js
waitingallday: Tried that and it didn't work :/

waitingallday:
I went back to the `ripple-nuxt-tide` rendered page and matched components and markup, the landing page is missing a couple of details that I've added here.

TODO: 
- [ ] Content rating form
- [ ] Footer, acknowledgement, links and copyright
- [ ] Cypress tests!

PR #1146, opened 2022-02-15
[INTCONTRIB-186] Added a honeypot form field component.

elNico:
## Motivation and Context

**JIRA issue:**  
https://salsadigital.atlassian.net/browse/HEALTHTRAN-441

This PR adds simple spam protection to every form generated via Tide/Drupal webform. There is no actual integration to any Tide settings, it simply adds a hidden frontend form field and if a bot fills it out, will not pass on any submission to the backend. It's been added as a simple Ripple form field component to make it possible to be easily used in coded schema-based form as well with minimal additions to the schema (currently processing in that case would need to be added individually).

## Changed

1. Added a honeypot form field component (`fieldRplhopo`), built on existing `fieldRplinput` with a few relevant properties set and CSS to make it invisible.
2. Added above field to the Storybook default form with a testing CSS class applied to make it visible.
3. Added above field to every form generated via Tide/Webform at the end, just before submit.
4. Added a check upon form submission (but after validation) to test whether the field contains any data and if so, to fake a successful submission.

### Screenshots

**Temporarily visible in Tide/Drupal webform example**

image

**Display in Storybook**

image

**Fake success message with `other` visual type applied to make manual testing easier**

image

## Types of changes

## Checklist


packages/ripple-nuxt-tide/modules/webform/components/TideForm.vue
```diff
@@ -73,22 +73,34 @@ export default {
...
-          response: {
-            status: 'success',
-            message: this.formData.messages.success || this.messages.success
+      if (formData.honeypot) {
+        // If the honeypot field has been filled, add delay and fake a success without further process.
+        setTimeout(() => {
+          this.formData.formState = {
+            response: {
+              status: 'other',
+              message: this.formData.messages.success || this.messages.success + ' (HP test)'
```
alan-cole, line 82: Should `status: other` and `(HP test)` be output here? Or this this left-over code from testing?

I'm assuming the success message shown to honeypot captured submissions should be identical to the real success output message.

packages/components/Molecules/Form/fields/fieldRplhopo.vue
```diff
@@ -0,0 +1,42 @@
...
+import { abstractField } from 'vue-form-generator'
+export default {
+  name: 'RplHopo',
+  components: {
+    RplInput
+  },
+  mixins: [abstractField],
+  mounted () {
+    this.schema.label ??= 'Important email'
+    this.schema.model ??= 'honeypot'
```
alan-cole, line 24: Just to be safe, has this been tested on IE11 to make sure the `??=` operator is being properly transpiled? If it has been tested and works, then all good 👍

alan-cole (changes requested):

packages/ripple-nuxt-tide/modules/webform/components/TideForm.vue
elNico: @alan-cole yes, that's what I originally had. But I wanted to make it easy to test without any coding and checked the output. Nothing should really alert a bot that this was a fake success message.

But I'd also be fine removing the differences. I'll see what @dylankelly thinks.

packages/components/Molecules/Form/fields/fieldRplhopo.vue
elNico: @alan-cole I will check browser support.

packages/ripple-nuxt-tide/modules/webform/components/TideForm.vue
elNico: Removed text, left `status: other` for now.

packages/components/Molecules/Form/fields/fieldRplhopo.vue
elNico: Not compatible, changed.

dylankelly:
@alan-cole I'm not sure why a seperate field is needed when you can add `styleClasses` to the existing `rplInput` field?

elNico:
@dylankelly yes, it is mainly a convenience and essentially a wrapper around `rpl-input`. However, there's a few more properties beyond the CSS that someone who would do this on a case-by-case basis may not add, but that could be helpful in some situations, like suitable defaults for autocomplete, tabindex etc.

But it would also be an aim to minimise work and standardise how to add a honeypot method to a form, ideally being as simple as adding `{ type: 'rplhopo' }` and be done with it. We could also add the actual processing to the demo form submit handler so it serves as an example, since hiding an input field is obviously not the full honeypot method.

I guess you could compare it with other existing convenience fields/groups in`ripple-nuxt-tide` webform that are in the end just a bunch of input fields for a specific purpose. And the patch we did for Health Translations, where this is from, did only that, adding a specific input field to all webform-generated forms.

I added it as a component in this PR to have something to easily add if outside of the Tide/webform context when you're handcrafting a form. But yes, convenience :)

elNico:
@dylankelly an alternative I could think of would be to abstract it further, like a base form property that simply can be set like `spamProtect: true` and then adding the necessary field and validation would happen automatically without having to think about honeypot fields/method etc...

elNico:
Ok @dylankelly, I will remove it. However, I'll be having 2 weeks leave, so I'll pick this up again when I'm back. We can just leave it in the PR.

packages/components/Molecules/Form/index.vue
```diff
@@ -372,6 +405,10 @@ $rpl-form-input-search-icon: url("data:image/svg+xml,%3Csvg width='16' height='1
     .field-wrap {
       order: 3;
     }
+
+    &.rpl-hidden {
```
alan-cole, line 409: We have .rpl-hidden styles defined in [visibility.scss](https://github.com/dpc-sdp/ripple/blob/develop/packages/components/Atoms/Global/scss/utilities/_visibility.scss#L5) which should allow this to be hidden. Is this just a fail-safe?

alan-cole (approved):
I prefer this implementation to the previous one as it encapsulates the feature into a single option, rather than splitting the implementation out over several files. Good stuff 👍

packages/components/Molecules/Form/index.vue
alan-cole: @dylankelly - Nic mentioned that this bit of specific styling was required as using `rpl-hidden` normally would be overridden by some component specific styling of `display: flex`.

Should `rpl-hidden` in visibility.scss use `!important` to ensure it's not overridden by component styles?

dylankelly: @alan-cole yeah its probably a use case where `!important` makes sense, happy for the change to be made in this PR.

elNico: Thanks @alan-cole and @dylankelly, I've added the above.

alan-cole (dismissed):

elNico:
@dylankelly could you pls review again as well? Hopefully this will be a good compromise, no field, but an easy way to enable spam protection in just Ripple forms without having to think much about implementation. Tide webform then simply makes use of it.

I don't really know what's going on with the tests. Any help would be great. Thanks!

dylankelly:
> 

The error in storybook is :
`Error: Evaluation failed: ReferenceError: spamProtect is not defined`

Can you please test in Storybook - `yarn run start:storybook`, I think the configuration in the story must be missing something, but I have not investigated. This change should be opt in for all forms, so need to make sure we are not making any breaking changes here @elNico .

elNico:
Hi @dylankelly, so the `spamProtect` property was set to `false` by default all along, so existing forms shouldn't really be impacted and yes, it's opt-in. But the build error in SB was due to the fact that both examples, demo form and submissions alerts, share the same `TemplateForm` constant as the form template.

https://github.com/dpc-sdp/ripple/pull/1146/files#diff-79c8f77ee966b1a0a8fd79bbcfb616e7b5db8ebd56a26544109d8ba81c733427L71

So for the demo form I believe I need to pass it on there, but then it's undefined for the alerts. Short of them having separate templates altogether, I've set it to false for alerts. If there's some other way, pls let me know.

I did then find another quite weird bug where the values of "Option button default" were not being set anymore (see screenshot below). It seemed to have to do with the sequence of when the on-the-fly honeypot field gets inserted into the fields array in `mounted ()`...since moving it in the earlier hook `beforeMount ()` still worked and fixes the error I went with that. 

image

dylankelly:
> 

@elNico that makes sense, thanks for resolving.

dylankelly (approved):

PR #1180, opened 2022-04-05
[SDPA-6095] - Remove rainbow string from promo card

ChrisCampbellCodes:
## Motivation and Context

**JIRA issue:** https://digital-engagement.atlassian.net/browse/SDPA-6095

## Changed

1. Removed colour stripe from the top of the card component variant called 'promo no image'

### Screenshots
image

## Types of changes

- [x] Improvement/refactoring (non-breaking change that doesn't add any features but makes things better)

## Checklist

### DIFF
--- a/packages/components/Molecules/Card/CardPromo.vue
+++ b/packages/components/Molecules/Card/CardPromo.vue
@@ -144,7 +144,7 @@ export default {
   $rpl-card-promo-border-radius: $rpl-card-border-radius !default;
   $rpl-card-promo-no-image-border: rpl_gradient('decorative_gradient') !default;
   $rpl-card-promo-no-image-border-height: rem(8px) !default;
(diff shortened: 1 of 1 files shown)


waitingallday (approved):

dylankelly (changes requested):
@ChrisCampbellCodes as discussed - can we add a prop to this component that conditionally adds the `-no-image` class? We can then pass it to the card in the props that get passed to it through `ripple-nuxt-tide`

ChrisCampbellCodes:
> ?

Hey @dylankelly ,

I think we need a different way of handling the variant because the 'no-image' class is still used, its styles are split up so in the PR it looks like I've removed all of its styling but it is still an existing variant. The 'promo cards' are the ones that would have the option so perhaps we could do a 'promo card - feature' variant which includes the rainbow? And regular 'promo cards' like we have now wont have the rainbow.

PR #1248, opened 2022-09-14
[SW-2210]wrong date content collection

yeniatencio:
## Motivation and Context

**JIRA issue:**https://digital-vic.atlassian.net/browse/SW-2210

## Changed

1. Adding field_news_date for content type News for the search results. 

## Types of changes

## Checklist

### DIFF
--- a/examples/reference/.env
+++ b/examples/reference/.env
@@ -1,5 +1,5 @@
 # Tide API
-CONTENT_API_SERVER=https://develop.content.reference.sdp.vic.gov.au/
+CONTENT_API_SERVER=https://nginx-php.pr-129.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au
 # oAuth client key is needed for Preview to work
 CONTENT_API_OAUTH_CLIENT_ID=dc881486-c14a-4b92-a0d0-e5dcd706f5ad
 
--- a/examples/reference/.env
+++ b/examples/reference/.env
@@ -1,5 +1,5 @@
 # Tide API
-CONTENT_API_SERVER=https://nginx-php.pr-132.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au
+CONTENT_API_SERVER=https://develop.content.reference.sdp.vic.gov.au/
 # oAuth client key is needed for Preview to work
 CONTENT_API_OAUTH_CLIENT_ID=dc881486-c14a-4b92-a0d0-e5dcd706f5ad
 
--- a/examples/reference/.env
+++ b/examples/reference/.env
@@ -1,5 +1,5 @@
 # Tide API
-CONTENT_API_SERVER=https://nginx-php.pr-129.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au
+CONTENT_API_SERVER=https://nginx-php.pr-132.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au
 # oAuth client key is needed for Preview to work
 CONTENT_API_OAUTH_CLIENT_ID=dc881486-c14a-4b92-a0d0-e5dcd706f5ad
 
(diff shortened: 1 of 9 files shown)


packages/ripple-nuxt-tide/modules/landing-page/lib/content-collection.js
```diff
@@ -992,13 +992,13 @@ module.exports = class ContentCollection {
...
     const link = this.getLocalisedLinkFromSource(_source)
-
+    const newsDate = _source.field_news_date
     switch (this.getDisplayResultComponentType()) {
       case 'search-result':
         mappedResult = {
           title: _source.title?.[0],
           link: link ? { linkText: link.domain + link.path, linkUrl: link.path } : null,
-          date: _source.created?.[0],
+          date: newsDate ? _source.field_news_date?.[0] : _source.created?.[0],
```
waitingallday, line 1001: You're defining `newsDate` above, so you can use it here as well eg.
```
date: newsDate ? newsDate[0] : _source.created?.[0],
```
Otherwise you could drop that const entirely:
```
date: _source.field_news_date ? _source.field_news_date[0] : _source.created?.[0],
```

examples/reference/.env
```diff
@@ -1,5 +1,5 @@
 # Tide API
-CONTENT_API_SERVER=https://develop.content.reference.sdp.vic.gov.au/
+CONTENT_API_SERVER=https://nginx-php.pr-132.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au
```
waitingallday, line 2: Can you revert this change back to develop.content

dylankelly, line 2: @yeniatencio please revert this change.

The better way to change the CONTENT_API_SERVER value for a PR branch is to use the lagoon CLI tool to change the value, that way you don't need to commit this file. 

See https://digital-vic.atlassian.net/wiki/spaces/DDS/pages/2050949826/How-to+update+an+environment+variable+using+Lagoon

Otherwise if its just for local development just either run the dev command with the value before it eg: `CONTENT_API_SERVER=https://nginx-php.pr-132.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au npm run dev` _OR_ export the value, eg: `export CONTENT_API_SERVER=https://nginx-php.pr-132.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au`

packages/ripple-nuxt-tide/modules/landing-page/lib/content-collection.js
dylankelly, line 1001: This doesn't read very well as is, instead of a ternary here this could be a [logical OR](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR) and use [optional chaining ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to guard against undefined values  

eg 
```
date: _source.field_news_date?.[0] || _source.created?.[0]
```
This reads as:
if `field_news_date` is defined and the first item in the array is not falsy, use that otherwise use the other value

Even better would be to [destructure](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#binding_and_assignment) `_source` for just the values needed, but thats unrelated to this PR.

dylankelly (changes requested):

packages/ripple-nuxt-tide/modules/landing-page/lib/content-collection.js
yeniatencio: Thank you @dylankelly. I have updated my pr for your review.

examples/reference/.env
yeniatencio: Done @waitingallday Ta

dylankelly (approved):

ChrisCampbellCodes (approved):

PR #1260, opened 2022-11-17
[DDS-1368] Removed reference site hostnames from ripple project.

nicksantamaria:
Removes reference site hostnames from .lagoon.yml so they can be added to the actual reference site.


dylankelly (approved):
👍

PR #1262, opened 2022-11-21
[SDPAP-6961] Added changes to capture custom confirmation message using token for webforms.

MdNadimHossain:
## Motivation and Context
The BE added a new feature where you can use the hidden field to create a confirmation message using token which will be sent by the Post Request Response to the FE. It adds the custom message in the notes field of response data. The reason it couldn't use the existing confirmation message field because that field doesn't know the token value which gets generated after the submission and the exisiting confirmation field sends out the field before form submission.

**JIRA issue:** https://digital-vic.atlassian.net/browse/SDPAP-6961

## Changed

1. Returned the response data in webform mixin and used it to show the custom confirmation message with token value that returned from the webform post response by the BE.

### Screenshots
Screen Shot 2022-11-21 at 9 57 23 am

## Types of changes

## Checklist

### DIFF
--- a/packages/ripple-nuxt-tide/modules/webform/components/TideContentRating.vue
+++ b/packages/ripple-nuxt-tide/modules/webform/components/TideContentRating.vue
@@ -150,13 +150,13 @@ export default {
           }
         }
       } else {
-        const res = await this.postForm(formId, formData)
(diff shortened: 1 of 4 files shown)


dylankelly (approved):
@MdNadimHossain LGTM

PR #1305, opened 2023-06-05
[SDPAP-7477]remove call to function getExposedItemsToLoadField

yeniatencio:
## Motivation and Context

This change is required to remove the items per page dropdown box where it gives the option to select the number of pages shown. This option is being added in BE content collection by selecting numbers of results shown. 3, 6, 9.

**JIRA issue:** https://digital-vic.atlassian.net/browse/SDPAP-7477

BE link: https://nginx-php.pr-272.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au/
FE link: https://app.pr-1305.ripple.sdp4.sdp.vic.gov.au/
## Changed

1. Remove call to function getExposedItemsToLoadField. 
2. Remove required src in responsiveImage.vue because when selecting layout grid view and card display style Thumbnail in BE for the content collection component and if any of the results has no image then it will throw an error of Missing required prop: src . There is already a validation on the template 
```
<template>
  <img v-if="src" class="rpl-responsive-image" :class="`rpl-responsive-image--fit-${fit}`" :alt="alt" :height="height" :width="width" ref="image" :src="src" :srcset="calcSrcSet" :sizes="calcSizes" :style="calcFocalPoint"  />
</template>

```

### Screenshots
Screenshot 2023-06-05 at 4 21 34 pm

## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/components/Atoms/ResponsiveImg/ResponsiveImg.vue
+++ b/packages/components/Atoms/ResponsiveImg/ResponsiveImg.vue
@@ -20,9 +20,7 @@ export default {
   mixins: [breakpoint],
   props: {
     src: {
-      type: String,
-      required: true
-    },
+      type: String    },
     alt: {
       type: String
     },
--- a/packages/components/Atoms/ResponsiveImg/ResponsiveImg.vue
+++ b/packages/components/Atoms/ResponsiveImg/ResponsiveImg.vue
@@ -20,7 +20,8 @@ export default {
   mixins: [breakpoint],
   props: {
     src: {
-      type: String
+      type: String,
+      required: true
     },
     alt: {
       type: String
--- a/packages/components/Atoms/ResponsiveImg/ResponsiveImg.vue
+++ b/packages/components/Atoms/ResponsiveImg/ResponsiveImg.vue
@@ -20,7 +20,8 @@ export default {
   mixins: [breakpoint],
   props: {
     src: {
-      type: String    },
+      type: String
+    },
     alt: {
       type: String
     },
--- a/packages/ripple-nuxt-tide/modules/landing-page/lib/content-collection.js
+++ b/packages/ripple-nuxt-tide/modules/landing-page/lib/content-collection.js
@@ -826,6 +826,7 @@ module.exports = class ContentCollection {
   getExposedControlFields () {
     const controls = [
       this.getExposedSortField()
+      this.getExposedItemsToLoadField()
     ]
     return controls.filter(item => item !== null)
   }
--- a/packages/ripple-nuxt-tide/modules/landing-page/lib/content-collection.js
+++ b/packages/ripple-nuxt-tide/modules/landing-page/lib/content-collection.js
@@ -826,7 +826,6 @@ module.exports = class ContentCollection {
   getExposedControlFields () {
     const controls = [
       this.getExposedSortField(),
-      this.getExposedItemsToLoadField()
     ]
     return controls.filter(item => item !== null)
   }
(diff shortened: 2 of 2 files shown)


packages/components/Atoms/ResponsiveImg/ResponsiveImg.vue
```diff
@@ -20,8 +20,7 @@ export default {
   mixins: [breakpoint],
   props: {
     src: {
-      type: String,
-      required: true
```
dylankelly, line 24: @yeniatencio this doesn't look relevant to the pr, can you explain why this is here?

yeniatencio: When selecting layout grid view and card display style Thumbnail in BE for the content collection component and if any of the results has no image then it will throw this error. 
Screenshot 2023-06-05 at 6 13 27 pm

dylankelly: it seems like the src should be required for the image, you've fixed the issue by removing the required prop but the ResponsiveImg component shouldn't be rendered at all. @lambry can I get you to look at this and provide some feedback?

alan-cole:
Hi @dylankelly, @yeniatencio - I just noticed this PR in passing. The change here will break places where an exposed items per page is required, like on https://www.health.vic.gov.au/search?q=eye.
If the purpose is to set the items per page, but without the user configurable option, then we shouldn't be using 
```
$config['interface']['display']['options']['itemsToLoad']['values']
```
from [here](https://github.com/dpc-sdp/tide_api/pull/108/files#diff-c5bea1becd4ebb9c10aba8223e1cc50b027d6050fdfffb76024ef352c57f1c20R1794) to set the value.

Instead, try using `internal.itemsToLoad` [see schema](https://github.com/salsadigitalauorg/sdp-content-collection-schema/blob/master/schema.js#L192):
```
$config['internal']['itemsToLoad']
```
This should be achievable in the back-end by updating the JSON configuration for the content collection, without needing this PR.

packages/components/Atoms/ResponsiveImg/ResponsiveImg.vue
lambry: Hey @dylankelly @yeniatencio, 

In this instance looks like the card component is being passed an empty object and we aren't checking this is valid, we just assume an object means it's a real image. We might need to update the below check to make sure the object actually has a source, or potentially update the mapping. This sounds like a separate ticket though. 

Cheers

REF: https://github.com/dpc-sdp/ripple/blob/develop/packages/components/Molecules/Card/mixins/card.js#L119

dylankelly:
@yeniatencio In talking this through with @alan-cole , I don't think we should proceed with this PR. It looks like there has been some confusion here. The required change to add a defined number of results is needed for Ripple 2 only. By removing the ability to change the number of results we would break existing functionality in Ripple 1 (See link Alan provided). 

@lambry we might need to pickup this ticket for Ripple 2. 

Sorry for the confusion.

yeniatencio:
> @yeniatencio In talking this through with @alan-cole , I don't think we should proceed with this PR. It looks like there has been some confusion here. The required change to add a defined number of results is needed for Ripple 2 only. By removing the ability to change the number of results we would break existing functionality in Ripple 1 (See link Alan provided).
> 
> @lambry we might need to pickup this ticket for Ripple 2.
> 
> Sorry for the confusion.

Hi @dylankelly , I was going to close this pr too and Yes agree with you. I have already deleted the variable from this pr that I was using with my BE pr.   Thanks guys.

PR #1317, opened 2023-08-31
[SDPAP-7863]added fix to have one publication author

yeniatencio:
## Motivation and Context

We are updating publication author field to be restricted to one value (organisation taxonomy)across all projects then we need to update it in the FE.

Currently there is also an issue for publication author not being displayed on SSP project and this is happening because this field is contemplating one value in the BE. 

**JIRA issue:** https://digital-vic.atlassian.net/browse/SDPAP-7863
BE ticket : https://digital-vic.atlassian.net/browse/SDPAP-7925
Ripple pr link: https://app.pr-1317.ripple.sdp4.sdp.vic.gov.au/
BE testing link: https://nginx-php.pr-323.content-reference-sdp-vic-gov-au.sdp4.sdp.vic.gov.au/
## Changed
1.  Updated the type string for author prop.
2. Changed to get the string value of author name from the field.
3. Added same change for the print page.

Tide_publication pr needs to be merged before to have this pr merged.

### Screenshots
Screenshot 2023-08-31 at 6 57 42 pm
Screenshot 2023-08-31 at 6 58 52 pm

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/components/Organisms/Publication/PublicationAuthorInformation.vue
+++ b/packages/components/Organisms/Publication/PublicationAuthorInformation.vue
@@ -5,6 +5,7 @@
 <script>
 import formatdate from '@dpc-sdp/ripple-global/mixins/formatdate'
 import RplDescriptionList from '@dpc-sdp/ripple-description-list'
+import pluralize from 'pluralize'
 
 export default {
   name: 'RplAuthorInformation',
@@ -22,7 +23,8 @@ export default {
     informationListing () {
       const terms = []
       if (this.author) {
-        terms.push({ term: 'Author', description: this.author })
+        const numberOfAuthors = (this.author.match(/,/g) || []).length + 1
+        terms.push({ term: pluralize('Author', numberOfAuthors), description: this.author })
       }
       if (this.date) {
         terms.push({ term: 'Date', description: this.formatDate(this.date, 'MMMM YYYY') })
--- a/packages/components/Organisms/Publication/PublicationAuthorInformation.vue
+++ b/packages/components/Organisms/Publication/PublicationAuthorInformation.vue
@@ -14,7 +14,7 @@ export default {
     RplDescriptionList
   },
   props: {
-    author: Array,
+    author: String,
     date: String,
     copyright: String,
     locale: { default: 'en-au', type: String }
@@ -22,8 +22,8 @@ export default {
   computed: {
     informationListing () {
       const terms = []
-      if (this.author && this.author.length > 0) {
-        terms.push({ term: pluralize('Author', this.author.length), description: this.author.join(', ') })
+      if (this.author) {
+        terms.push({ term: 'Author', description: this.author })
       }
       if (this.date) {
         terms.push({ term: 'Date', description: this.formatDate(this.date, 'MMMM YYYY') })
--- a/packages/components/Organisms/Publication/PublicationAuthorInformation.vue
+++ b/packages/components/Organisms/Publication/PublicationAuthorInformation.vue
@@ -14,7 +14,7 @@ export default {
     RplDescriptionList
   },
   props: {
-    author: String,
+    author: Array,
     date: String,
     copyright: String,
     locale: { default: 'en-au', type: String }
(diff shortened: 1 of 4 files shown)


lambry (changes requested):
Hey @yeniatencio, 

One super minor thing, can you update this publication storybook story as well.

Screenshot 2023-09-05 at 2 25 56 pm

Is there a chance some sites will return an array of authors and others will return the single value while the backends are being updated? Just wondering if we would need to support both.

Cheers

yeniatencio:
> Hey @yeniatencio,
> 
> One super minor thing, can you update this publication storybook story as well.
> 
> Screenshot 2023-09-05 at 2 25 56 pm
> Is there a chance some sites will return an array of authors and others will return the single value while the backends are being updated? Just wondering if we would need to support both.
> 
> Cheers

Hey @lambry , I have created a pr in core to have the publication author field a single value in all BEs.  So no chance to have more than 1 single value for Publication author. 
PS, I have updated the storybook, Thanks

packages/ripple-nuxt-tide/modules/publication/pages/PrintAll.vue
```diff
@@ -105,7 +105,7 @@ export default {
     publishingInfo () {
       if (this.publication.type === 'node--publication') {
         const authorsName = this.publication.field_publication_authors ? this.publication.field_publication_authors : ''
-        const author = authorsName && authorsName.length > 0 ? authorsName.map((authors) => authors.name) : []
+        const author = authorsName ? authorsName.name : ''
```
dylankelly, line 108: @yeniatencio I think we should support both an array and a single here so we dont break this behaviour anywhere?

dylankelly (changes requested):

packages/ripple-nuxt-tide/modules/publication/pages/PrintAll.vue
dylankelly: See https://github.com/dpc-sdp/ripple-framework/pull/837/files#diff-0b5c2b5d5780442faddb8f6ba0cba008efbdaa0710114f62735bde65f8db61a9R45

yeniatencio: @dylankelly , I have updated my pr for your review. Thank you.

packages/components/Organisms/Publication/PublicationAuthorInformation.vue
```diff
@@ -14,16 +14,17 @@ export default {
...
     copyright: String,
     locale: { default: 'en-au', type: String }
   },
   computed: {
     informationListing () {
       const terms = []
-      if (this.author && this.author.length > 0) {
-        terms.push({ term: pluralize('Author', this.author.length), description: this.author.join(', ') })
+      if (this.author) {
+        const numberOfAuthors = (this.author.match(/,/g) || []).length + 1
```
lambry, line 26: Hey @yeniatencio, 

Going by the storybook example authors can be departments i.e. "Department of Premier and Cabinet", this means an author name could have a comma in it like "Department of Families, Fairness and Housing", which would lead to the incorrect number of authors i.e. pluralization. 

We could do this the opposite way and revert the PublicationAuthorInformation.vue so it accepts an array again then in the author mapping stuff make sure author is always an array.

```
const author = Array.isArray(authorsName) ? authorsName.map((author) => author.name) : [authorsName.name]
```

yeniatencio: Good call @lambry . I have updated my pr for your review. Thanks

lambry (approved):

yeniatencio:
Hey @dylankelly , can you please re-review, i have now changed the code to handle the case of older tide with an array, and newer tide with a single value which was David's feedback.

dylankelly (approved):
Thanks @yeniatencio - appreciate you adding this to prevent any surprises.

yeniatencio:
> Thanks @yeniatencio - appreciate you adding this to prevent any surprises.

Thank to you @dylankelly . Good thing to anticipate that scenario.

PR #1320, opened 2023-10-13
Updated nightwatch workflow to e2e

MarkCalvert:


.github/workflows/e2e.yml
```diff
@@ -1,4 +1,4 @@
-name: dpc-sdp/ripple/nightwatch
+name: dpc-sdp/ripple/e2e
```
MarkCalvert, line 1: @tim-yao Do we need to update all input values for test_type from `test_type: ':nightwatch:'` to `test_type: ':e2e:'` as well?

tim-yao: > @tim-yao Do we need to update all input values for test_type from `test_type: ':nightwatch:'` to `test_type: ':e2e:'` as well?

That is a good suggestion, please do. Cheers!

tim-yao (dismissed):

MarkCalvert:
@tim-yao I have updated all references from nightwatch to e2e. 
Ready for your review.

tim-yao (approved):

PR #1358, opened 2026-03-01
feat(@dpc-sdp/ripple-ui-forms): adds ability to control form step progression [SD-1587]

dylankelly:
**Issue**: https://digital-vic.atlassian.net/browse/SD-1587

### What I did

- Adds ability to pass down [beforeStepChange](https://formkit.com/plugins/multi-step#the-beforestepchange-event) function. This allows step progression to be controlled based upon form values, etc. 
- Adds multi step form story, including example beforeStepChange function to demonstrate branching logic flow
- Adds parentStep property to RplFormStep. This will make the step not appear in the progress bar, and hte parent step will appear as the active step. This can be used to have child steps that do not appear in the navigation, which is needed for branching logic.
- Fixes some imports for missing components (RplProgress, RplSummaryList, RplFormAlert)
- Adds a basic multi step component test
 

### How to test

- See multi step form navigation in storybook. Use the value "test" in the organisation field to proceed to the eligible step, all other values will say not eligible 
- 

#### For all PR's

- [ ] I've added relevant changes to the project Readme if needed
- [ ] I've updated the documentation site as needed
- [ ] I have added tests to cover my changes (if not applicable, please state why in a comment)

#### For new UI components only

- [ ] I have added a storybook story covering all variants
- [ ] I have checked a11y tab in storybook passes
- [ ] I have added cypress component tests (if the component is interactive)
- [ ] Any events are emitted on the event bus using `emitRplEvent`


packages/ripple-ui-forms/src/components/RplForm/RplForm.vue
```diff
@@ -421,7 +421,7 @@ const handleStepChange = async ({
 
   // Get the current steps errors when it's invalid, and we're trying to proceed
   if (!currentStep.isValid && forwards) {
-    cachedErrors.value = getErrorMessages(getNode(currentStep.id))
+    cachedErrors.value = getErrorMessages(currentStep.node)
```
lambry: 👍

dylankelly:
> Do you think this same function can be used to determine if the 'next' button should be disabled/hidden etc?

hmmm, I dont think so because it runs on step change and the button is already rendered. We can easily add a props at the step level to disable the next / prev button, might add it in a seperate PR. I dont think we need to make it conditional at this point.

waitingallday:
Will need to revisit this later

